PR #30, opened 2019-04-09
ifgram_inversion: bug fixed while runing in command line

yunjunz:


2gotgrossman:
Did you mean to make changes to `save_roipac.py` and `plot.py`?

2gotgrossman:
If you didn't mean to, that could be a good reason to follow the standard Pull Request process 😄

yunjunz:
They are not really related with ifgram_inversion. I used to attached them together and wrote another line of message, to avoid too many commits. Is it recommend to have them pushed separated?

PR #32, opened 2019-04-09
Added Dask config defaults & fixed #31

### LINKED ISSUE #31: Dask Throws Exception Once in a While due to `gamma` import
This Exception used to be thrown once in a while when running with Dask, but now it is thrown consistently (the last 4 times I have run the code with Dask).

@yunjunz , you mentioned that we could just add the `gamma` function to our code and that would solve the problem. Is that still possible? 

```
  File "pysar/ifgram_inversion.py", line 1220, in <module>
    main()
  File "pysar/ifgram_inversion.py", line 1211, in main
    ifgram_inversion(inps.ifgramStackFile, inps)
  File "pysar/ifgram_inversion.py", line 1117, in ifgram_inversion
    for future, result in as_completed(futures, with_results=True):
  File "/nethome/dwg11/anaconda2/envs/pysar_new/lib/python3.6/site-packages/distributed/client.py", line 3858, in __next__
    return self._get_and_raise()
  File "/nethome/dwg11/anaconda2/envs/pysar_new/lib/python3.6/site-packages/distributed/client.py", line 3849, in _get_and_raise
    six.reraise(*result)
  File "/nethome/dwg11/anaconda2/envs/pysar_new/lib/python3.6/site-packages/six.py", line 692, in reraise
    raise value.with_traceback(tb)
  File "/nethome/dwg11/anaconda2/envs/pysar_new/lib/python3.6/site-packages/distributed/protocol/pickle.py", line 59, in loads
    return pickle.loads(x)
  File "/nethome/dwg11/anaconda2/envs/pysar_new/lib/python3.6/site-packages/numpy/core/__init__.py", line 149, in _ufunc_reconstruct
    return getattr(mod, name)
AttributeError: module '__mp_main__' has no attribute 'gamma'
```

2gotgrossman:
This commit is larger than I'd like it to be but it includes the following changes:
1. Creation and use of Dask configuration using [Dask's configuration setup](http://jobqueue.dask.org/en/latest/configuration-setup.html)
2. A fix of #31, where Dask workers cannot find the `gamma` function and therefore throw an exception
3. Moving the imports for Dask to the top of `ifgram_inversion.p`
4. Updates to the documentation
5. Having the code where `inps.parallel` is False be before the parallel case. I think this makes the code more readable

@yunjunz could you please test to make sure the `inps.parallel = False`  case works?

Also, once this Pull Request is merged, you should be able to run Dask on Pegasus!!!

### DIFF
--- a/pysar/ifgram_inversion.py
+++ b/pysar/ifgram_inversion.py
@@ -10,6 +10,7 @@
 
 import os
 import re
+import sys
 import time
 import argparse
 import h5py
@@ -18,6 +19,12 @@ from scipy import linalg   # more effieint than numpy.linalg
 from scipy.special import gamma
 from pysar.objects import ifgramStack, timeseries
 from pysar.utils import readfile, writefile, ptime, utils as ut
+
+# Imports for parallel execution
+from dask.distributed import Client, as_completed
+# dask_jobqueue is needed for HPC.
+# PBSCluster (similar to LSFCluster) should also work out of the box
+from dask_jobqueue import LSFCluster
 
 
 # key configuration parameter name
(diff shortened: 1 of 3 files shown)


pysar/ifgram_inversion.py
```diff
@@ -583,15 +611,15 @@ def subsplit_boxes_by_num_workers(box, num_workers, dimension='y'):
 
     if dimension == 'y':
         y_diff = y1 - y0
-        for i in range(num_workers):
-            start = (i * y_diff) // num_workers
-            end = ((i + 1) * y_diff) // num_workers if i + 1 != num_workers else y_diff
+        for i in range(number_of_splits):
+            start = (i * y_diff) // number_of_splits
+            end = ((i + 1) * y_diff) // number_of_splits if i + 1 != number_of_splits else y_diff
```
Ovec8hkin: Could you add some comments on what the `start` and `end` computations are? They are not immediately clear to me.

2gotgrossman: Hmm, let's talk about this in person. I had a hard time coming up with a comment so I'm wondering when I explain it to you in person, which part was the missing thing. I can then include that in the comment

2gotgrossman:
@yunjunz I made some updates. Let me know your thoughts.

yunjunz:
Looks all good to me, @2gotgrossman. Thank you for making the parallel processing happen! I shortened the `workers-num` option, if you think it's okay, please squash and merge it.

2gotgrossman:
> Looks all good to me, @2gotgrossman. Thank you for making the parallel processing happen! I shortened the `workers-num` option, if you think it's okay, please squash and merge it.

Hey @yunjunz , good call at shortening the parameter. I might suggest to include `parallel` in the name, so it's obvious what the parameter is for. Like `parallel-workers-num` (or why not `parallel-num-workers`?)

PR #35, opened 2019-04-14
ifgram_inversion: import dask module for --parallel option only

yunjunz:
This PR move the `dask` modules import from the top of the script to `cmd_line_parse()` and execute the import when `inps.parallel==True` only, so that users who do not want to parallel option can still run it without installing dask.

### DIFF
--- a/pysar/ifgram_inversion.py
+++ b/pysar/ifgram_inversion.py
@@ -20,12 +20,6 @@ from scipy import linalg   # more effieint than numpy.linalg
 from pysar.objects import ifgramStack, timeseries
 from pysar.utils import readfile, writefile, ptime, utils as ut
 
-# Imports for parallel execution
-from dask.distributed import Client, as_completed
-# dask_jobqueue is needed for HPC.
-# PBSCluster (similar to LSFCluster) should also work out of the box
-from dask_jobqueue import LSFCluster
-
 
 # key configuration parameter name
 key_prefix = 'pysar.networkInversion.'
@@ -129,7 +123,7 @@ def create_parser():
                         help='max number of data (= ifgram_num * num_row * num_col) to read per loop\n' +
                         'default: 0.2 G; adjust it according to your computer memory.')
 
-    par = parser.add_argument_group('parallel', 'parallel processing configuration')
+    par = parser.add_argument_group('parallel', 'parallel processing configuration for Dask')
     par.add_argument('--parallel', dest='parallel', action='store_true',
                      help='Enable parallel processing for the pixelwise weighted inversion.')
     par.add_argument('--parallel-workers-num','--par-workers-num','--parallel-num', dest='num_workers', type=int,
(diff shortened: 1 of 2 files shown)


yunjunz:
Hi @2gotgrossman, I think it's important to be able to **not require `dask` installed** if the user did not use `--parallel` option; otherwise, all the current users who have not installed it will not be able to run the code.

One way to do it is to move the module import inside the main function as I push here; if you have a more elegant solution, please let me know.

2gotgrossman:
Hey @yunjunz, since installing Dask is already in the `requirements.txt` and the `conda.txt` files, they should already have it installed, so whether they import it or not shouldn't matter. The standard is to keep `import`s at the top of a file and would decrease readability and increase complexity of imports.

Besides the convention issues and it feeling wrong, I could see why it would be useful not to import it unless it's used.

Also, I didn't mean to reference #37 here. I meant to reference. I meant to reference it in #36

yunjunz:
I agree that it's not the convention. In practice, non-expert users don't re-install the Python dependencies as frequently as developers do; however, using the latest feature or bugs fix is still desired. Therefore, it will be confusing when one updated the code with `git pull` and it runs into error with old settings. For example, ISCE uses `CUDA` to for GPU computing, `RelaxIV` for 2 stage phase unwrapping, and I have not used any of them by far (installing them is difficult, but it's not relevant in our case here).

A compromised way could be import at the top and assigning a global variable for the import status, as shown below. I don't really like it because it does not use the native `ImportError` message. Let me know what do you think please.
https://github.com/yunjunz/pyaps3/blob/master/rdrcoord.py#L21

2gotgrossman:
I think the workaround that developers use is that they version code. The way non developers should use your code is that they download your code once and use it. If they want to update the code they use, then they re-download. But if they install your code via `pip`, this should all be handled for them. If they install your code from source, I think they won't update their version frequently.

Thinking about how to do code versioning would be good. Maybe release a new version every time there's a new import? Possibly something to look into there.

The only exception it looks like there is from this [StackOverflow thread on middle-of-script imports](https://stackoverflow.com/questions/1188640/good-or-bad-practice-in-python-import-in-the-middle-of-a-file) is if there's a possibility of circular imports. That isn't hte case for us so I imagine we can come up with an appropriate workaround.

yunjunz:
Versioning is definitely a way, will do it soon.

I am all for the code readability, but code that can be easily run has higher priority; we can clean it later. 

Below is another reason.
> If the imported module is infrequently used and the import is expensive, the in-the-middle-import is OK. Otherwise, is it wise to follow Alex Martelli's suggestion.
https://stackoverflow.com/questions/1188640/good-or-bad-practice-in-python-import-in-the-middle-of-a-file

falkamelung:
I agree with David. If somebody wants to figure out dask this adds to the confusion

Sent from my IPhone.

On Apr 15, 2019, at 6:08 PM, Zhang Yunjun <notifications@github.com<mailto:notifications@github.com>> wrote:


Versioning is definitely a way, will do it soon.

I am all for the code readability, but code that can be easily run has higher priority; we can clean it later.

Below is another reason.

If the imported module is infrequently used and the import is expensive, the in-the-middle-import is OK. Otherwise, is it wise to follow Alex Martelli's suggestion.
https://stackoverflow.com/questions/1188640/good-or-bad-practice-in-python-import-in-the-middle-of-a-file<https://nam01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fstackoverflow.com%2Fquestions%2F1188640%2Fgood-or-bad-practice-in-python-import-in-the-middle-of-a-file&data=02%7C01%7Cfamelung%40rsmas.miami.edu%7Cd66da3789a464fd7bdeb08d6c1eedbd1%7C2a144b72f23942d48c0e6f0f17c48e33%7C0%7C0%7C636909628972796971&sdata=ecagp8%2BH%2FinBzL6p5u16lLO5sFfrrBOEWVZcPvbrYSE%3D&reserved=0>

yunjunz:
Okay. Thank you for your opinion @falkamelung and for your patient discussion @2gotgrossman .

PR #46, opened 2019-04-25
pysarApp: expose more parameters for empirical tropospheric delay correction

leixie-steven:
Hi Yunjun,

Issue: The pysarApp_template has the option "pysar.troposphericDelay.looks" for the multi-look in empirical troposphere correction. However, this parameter is not been passed when you call the method.

Will you consider to add the threshold setting option for the empirical troposphere correction since you have already implemented in the method? Or just leave it for the user-defined pipeline?

Regards,
Lei

### DIFF
--- a/pysar/defaults/pysarApp.cfg
+++ b/pysar/defaults/pysarApp.cfg
@@ -55,12 +55,12 @@ pysar.networkInversion.numWorker        = 40
 
 
 ########## Tropospheric Delay Correction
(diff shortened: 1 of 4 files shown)


leixie-steven:
Hi Yunjun,
It makes clearer using your suggested parameter name. I have updated it.
Best regards,
Lei

yunjunz:
Looks all good to me. Thank you Lei.

Yunjun

PR #58, opened 2019-05-03
geocode: add OMP_NUM_THREADS control

yunjunz:
+ add `geocode/check_num_processor()` to check `OMP_NUM_THREADS` if it's defined in environment variable and use it if `--nprocs` is not specified.

+ set default `--nprocs` to 1 because 1) bonding between `pykdtree` and `openmp` is not clear to install/setup yet and 2) the speedup is negligible for macport install and slows down for conda install

+ change default nprocs value to 1 in objects/resample.py

+ detect lookup table file in current directory besides the INPUTS directory.

### DIFF
--- a/pysar/geocode.py
+++ b/pysar/geocode.py
@@ -70,8 +70,8 @@ def create_parser():
     parser.add_argument('--fill', dest='fillValue', type=float, default=np.nan,
                         help='Value used for points outside of the interpolation domain.\n' +
                              'Default: np.nan')
-    parser.add_argument('-n','--nprocs', dest='nprocs', type=int,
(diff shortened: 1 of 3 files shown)


yunjunz:
Hi @hfattahi, I have cleared the OMP_NUM_THREADS problem: now only one processor will be used by default, unless it's manually specified while calling geocode.py.

hfattahi:
It's a closed PR :). Thanks Yunjun. With these changes, if one asks for let say 4 processor, have you checked if it really uses 4 ?

yunjunz:
I have checked it. For conda installed environment: yes, it actually uses 4; for macports installed environment: no, it does not because macport does not have openmp support.

PR #76, opened 2019-06-06
ARIA standard products

hfattahi:
This PR adds a stand-alone script "prep_aria.py" which creates interferogram and geometry HDF5 files from ARIA standard products. prep_aria.py is not integrated to load_data.py in mintpy. One needs to first run ARIA-tools (will be available soon) to create required files which perp_aria.py works with. 
A minor bug is also fixed in which checking lookup files is skipped for geocoded data.

### DIFF
--- a/mintpy/prep_aria.py
+++ b/mintpy/prep_aria.py
@@ -1,0 +1,217 @@
+#!/usr/bin/env python3
+import os
+import gdal
+import h5py
+import argparse
+import numpy as np
+from mintpy.utils import ptime
+
+EXAMPLE = """ example:
+
+prep_aria.py -w mintpy_SanFran -s stack/stack/ -i stack/incidenceAngle/20150605_20150512.vrt -b stack/bPerpendicular/ 
+
+"""
+
+def create_parser():
+    """Command line parser."""
+    parser = argparse.ArgumentParser(description='Prepare ARIA processed products for MintPy.',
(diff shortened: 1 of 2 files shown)


hfattahi:
@ehavazli I think in the ARIA products, the refernce date is always the second date and the following date is the first date. I think MintPy expects the opposite. I therefore swapped the dates within prep_aria.py and multiplied baseline by -1. Maybe we should do the same for phase?

ehavazli:
@hfattahi I believe the products have the right phase information but I am not sure how mintpy treats it. I have an example area to check and I can share the result with you.

hfattahi:
It's not about right or wrong phase information. It's about the choice of "sign". In Mintpy time-series outputs, subsidence comes out as negative values and uplift as positive. If you have a case study that you know what ground displacement expect, then we can fix the sign in prep_aria.py if a fix is needed.

ehavazli:
right as in the proper sign however the results that came out over Norfolk is in the wrong direction, therefore there should be a fix. The problem I had with this is there are a few spots in my track which seem to have the correct direction of deformation but I just confirmed with David's intern about Norfolk so the phase should also be multiplied with -1.

hfattahi:
@ehavazli Thanks Emre. I just issued a PR which should fix the sign. BTW, I think we should open issues for these kind of discussions instead of commenting on closed PRs. Maybe next time :)

PR #91, opened 2019-06-13
fix the sign of unwrapped phase in prep_aria.py

hfattahi:
This PR multiplies the sign of ARIA products unwrapped phase by -1 to be consistent with what MintPy expects. This change in sign is because the reference date in ARIA products are the second date and the following dates are the first date. This order has been swapped in prep_aria.py to be consistent with Mintpy.


hfattahi:
For some reason the PR shows the commits in the previous PR. What is new in this PR is just this one:

https://github.com/insarlab/MintPy/pull/91/commits/a78b1f306d691e86b3596bc993f08ad5af516279

yunjunz:
Hi @hfattahi, I guess this is caused by not rebasing on top of the upstream/master, which is described in details below. 

https://mintpy.readthedocs.io/en/latest/CONTRIBUTING/#1_working_with_a_feature_branch

I disabled the `merge commits` option and allow only `squash merging` and `rebase merging` only on the `master` branch to ensure a more readable commit history.

yunjunz:
#76

PR #101, opened 2019-06-19
Init of basic snap (BEAM-DIMAP) ingest functionality

andretheronsa:
Add prep_snap which is based on prep_isce and prep_roipac that extracts metadata from interferogram/coherence/unwrapped/elevation .dim (BEAM-DIMAP format) files produced by ESA SNAP and writes a RSC file in ROIPAC format in the same folder as the .dim file. HEIGHT and EARTH_RADIUS are extracted from orbit parameters using code adapted and simplified from ISCE2.
Furthermore snap is added as processor in load_data.
This update does not finalize SNAP integration - the next issue will likely be to read the data in the .data folder using mintpy.utils.readfile since the data (in envi format) is in a .data folder one level deeper than the asociated .dim / rsc files.

### DIFF
--- a/mintpy/load_data.py
+++ b/mintpy/load_data.py
@@ -52,36 +52,42 @@ DEFAULT_TEMPLATE = """template:
            auto_path.gammaAutoPath)
 
 TEMPLATE = """template:
-########## 1. Load Data (--load to exit after this step)
+########## 1. Load Data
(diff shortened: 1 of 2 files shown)


yunjunz:
Thank you @andretheronsa!

Issue #100, opened 2019-06-19
tracking no-data for GDAL datasets in ARIA products


dbekaert:
@hfattahi think we might need to account for the no-data values in the connected component data for ARIA, perhaps in general as well for unwrapped etc. @bbuzz31 noticed some issue when using the unwrap correction using connected component to complain about negative value.

For example, zero is a connected component and therefore we set no-date value -1 in the ARIA product. Suggest to use something like the following to find the no-data values, and then set it to a value MintPy knows to ignore? Not sure if this would need to be 0 or a NaN or so?

For loop on line 155 in **prep_aria.py** each time when a band is loaded.
```
   bnd = dsUnw.GetRasterBand(ii+1)
   try:
        NoData = bnd.GetNoDataValue()
   except:
        NoData = None
```

hfattahi:
@dbekaert Yes I agree that we should better track NoData values. I believe Mintpy currently considers zero unwrapped phase as NoData and consistent with isce2 the connected component label of "0" is considered invalid. @yunjunz please correct me if I'm wrong.

In the meantime I think we can just shift the ARIA connected components by "1" while we load them to Mintpy and that should allow to use the unwrapping error correction code without modifications. something like this in prep_aria.py lines 126-127 :

```
bnd = dsComp.GetRasterBand(ii+1)
h5["connectComponent"][ii,:,:] = bnd.ReadAsArray() + 1
```
@dbekaert @bbuzz31 If you confirm this temporary fix is useful, we can commit it.

yunjunz:
> I believe Mintpy currently considers zero unwrapped phase as NoData and consistent with isce2 the connected component label of "0" is considered invalid. @yunjunz please correct me if I'm wrong.

This is correct @hfattahi.

dbekaert:
@hfattahi this would make connected component 0 be 1, and affect the current MintPy masking based on connected component 0. The no-data value for the connected component (i.e. currently -1) corresponds to the data outside the swath in the geo-coded space. Suggest to not shift the connected component, but instead put anything smaller than 0 to 0 as well. We are using no-data of 0 for unwrapped phase, so the current approach should be ok for now. 


```
bnd = dsComp.GetRasterBand(ii+1)
raster = bnd.ReadAsArray()
raster[raster<0]=0
h5["connectComponent"][ii,:,:] = raster
```

[...]

PR #103, opened 2019-06-19
fix bug of data type while reading isce .conncomp

yunjunz:
utils/readfile: fix bug while converting byte data type for ISCE file

objects/stackDict: change default connectComponent data type from bool to byte (int8) while loading from binary files

update comments accordingly


yunjunz:
#100

PR #118, opened 2019-07-12
add DEM file for ARIA

ehavazli:
- added the option to load DEM file

### DIFF
--- a/mintpy/prep_aria.py
+++ b/mintpy/prep_aria.py
@@ -5,9 +5,11 @@ import h5py
 import argparse
 import numpy as np
 from mintpy.utils import ptime
+from skimage.transform import resize
+
 
 EXAMPLE = """example:
-  prep_aria.py -w mintpy_SanFran -s stack/stack/ -i stack/incidenceAngle/20150605_20150512.vrt
+  prep_aria.py -w mintpy_SanFran -s stack/ -i incidenceAngle/20150605_20150512.vrt -d SRTM_3arcsec.dem
 """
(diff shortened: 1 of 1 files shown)


yunjunz:
> does this DEM neet to be at the grid of the stack?

The loaded DEM data should be in the same spatial coverage and spatial resolution as the rest data. 

@ehavazli, did you consider the different spatial coverage between the DEM file and the other data? `resize()` is convenient to resample the data, once it's cropped to the same spatial coverage.

sssangha:
@dbekaert when the DEM is loaded into memory it is cropped to the final bbox polygon. I thought we had decided not to write this out as a physical file though

dbekaert:
@sssangha  i think we should consider adding the option to extract the actual DEM on the IFG grid as well as part of ARIA-tools. This can be on user-request.

I see other users also needing this for options other than mintpy, that have their own TS approach or who want to do a simple tropo-phase correction etc

ehavazli:
> @ehavazli, did you consider the different spatial coverage between the DEM file and the other data? 

no, the code directly takes the given DEM file and only resize it to the length and width of the data. No cropping is done.

dbekaert:
Both the DEM and aria products are at 3arsec sampling. A difference in width and length implies the bounds are different. Added a ticket on ARIAtools (https://github.com/aria-tools/ARIA-tools/issues/64#issue-467700465) to add in option for extracting IFGdem. That should be the DEM you are looking for.

Issue #105, opened 2019-06-19
SNAP geocoded products support


yunjunz:
Hi @andretheronsa, thank you for adding `prep_snap.py` #101! Below are things that I can think of to make the SNAP+MintPy workflow as smooth as the ISCE+MintPy workflow.

+ [x] 1. Is it possible to add `interferograms` as the root directory of all interferograms? An example is shown below. This is preferred for a more simple root directory, so that `interferograms` folder can be in the same level as `mintpy` folder.

```
/interferograms
    /20190427_20190602
        /20190427_20190602_coh_tc.data
            /coh.img                                    #mintpy.load.corFile
            /coh.hdr
        20190427_20190602_coh_tc.dim
        /20190427_20190602_filt_int_sub_tc .data
            /filt_int.img                                #mintpy.load.intFile
            /filt_int.hdr
        20190427_20190602_filt_int_sub_tc.dim
        /20190427_20190602_unw_tc .data
            /filt_unw.img                             #mintpy.load.unwFile
            /filt_unw.hdr
        20190427_20190602_unw_tc.dim
    ...
/dem_tc.data 
    /dem.img                                         #mintpy.load.demFile
    /dem.hdr
dem_tc.dim
/mintpy
```

Once the question above is settled, `prep_snap` can be adjusted to run the complex metadata extraction function (`utm_dim_to_rsc()`) once and save the result into a simple text file, like `data.rsc` from prep_isce.py; and for the rest files, the code can search for `data.rsc` and used directly instead of calling `utm_dim_to_rsc()` again.

+ [x] 2. Could you write a document of your SNAP workflow, if there are more details than your comment inside prep_snap, on the mintpy's [wiki](https://github.com/insarlab/MintPy/wiki) or somewhere else you prefer? Other users who are interested can go there for detailed instruction to prepare SNAP products.

yunjunz:
@andretheronsa, if you have an updated file structure of the dataset that you uploaded to box, please let me know. I will use your exact file structure and move forward to test load_data.py and smallbaselineApp.py for it.

andretheronsa:
@yunjunz
1) I have uploaded the test data set with an updated structure to box.
1.1) Is there a specific reason _utm_dim_to_rsc()_ should only run once? I expect all the products to contain all the metadata and it runs within a second or two in my machine. At least for testing I propose we leave it as is.
[...]

PR #125, opened 2019-07-17
load_data: temp fix the file size inconsistency for SNAP products

yunjunz:
Add check_files_size() to check the file size of a list of files.

Adjust read_subset_box() to check the bounding box if 1) no input subset and 2) processor is snap.


yunjunz:
#105

PR #126, opened 2019-07-17
Initial Docker image with config file

andretheronsa:
Added instructions to run the Docker container from Dockerhub to Installation.md

Also added a root level docker directory for the dockerfile. Please review whether this is the right place, or even necessary. In this way anyone can inspect or modify the dockerfile and build their own container instead of using the prebuilt one on dockerhub.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -45,9 +45,7 @@ WORKDIR /home/python
 RUN git clone https://github.com/insarlab/MintPy.git
 
 # Pull and config PyAPS
-RUN mkdir -p PyAPS && \
-    cd PyAPS && \
-    git clone https://github.com/yunjunz/pyaps3.git
+RUN git clone https://github.com/yunjunz/pyaps3.git PyAPS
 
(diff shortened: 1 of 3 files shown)


yunjunz:
I moved the instruction of running docker to docker/README.md, following isce2 structure.

yunjunz:
Thank you for the commit @andretheronsa!

`Codacy` is suggesting to use `SHELL ["/bin/bash", "-c"]` to replace `RUN ln -sfv /bin/bash /bin/sh` to change the default shell, could you test if this works?

https://app.codacy.com/app/insarlab/MintPy/pullRequest?prid=3868908

andretheronsa:
@yunjunz I don't expect that to work. I believe SHELL only changes the default shell during the docker build time for commands you want to RUN in the dockerfile.
We need to change the shell 'permanently' in the actual container to ensure the .sh scripts get invoked correctly.

yunjunz:
@andretheronsa Thank you for the clarification.

Issue #83, opened 2019-06-08
Possibly add python>=3.6 to docs/conda.txt


piyushrpt:
This is related to #76 

When combining multiple requirements.txt (conda.txt for MintPy) to use gdal  3.0 features  with MintPy, conda's automatic resolution downgrades the python version to 2.7 to satisfy all the requirements. Adding python>=3.6 to conda.txt guarantees that a python3 environment is setup and that python doesnt get accidentally downgraded.

It would help to include this in the requirements file to allow for setup of environments where multiple tools can operate in the same env by just combining the requirements.txt files.

 For reference adding these 3 lines to conda.txt allows us to interoperate with gdal3.0
```
gdal>=3.0
libgdal
python>=3.6
```

yunjunz:
Hi @piyushrpt, I have added the 3 lines you suggested to `conda.txt`. Thank you for the recommendation.

piyushrpt:
@yunjunz Thanks for the update

piyushrpt:
Sorry for reopening the issue. Wanted to add information on conflict between proj.4 (version 6) that is needed for GDAL3 and matplotlib.basemap which needs the older version of (proj.4). This can be overcome by pointing PROJ_LIB to data directory provided by pyproj

```
export PROJ_LIB=`python3 -c "import pyproj; print(pyproj.pyproj_datadir)"`
```

piyushrpt:
There exists a minor conflict in handling the PROJ_LIB environment variable by the different packages. 

- For GDAL3, this needs to point to PROJ v6 which is in share/proj. This is automatically done by conda. 
- For basemap to continue to work with older PROJ conventions, this needs to point to 
```
export PROJ_LIB=`python3 -c "import pyproj; print(pyproj.pyproj_datadir)"`
```

For now one might have to toggle between the 2 settings for HDF5 preparation / visualization stages for #76

yunjunz:
Thank you @piyushrpt for the suggestion, I have updated the installation note accordingly.

hfattahi:
@yunjunz For now let's remove the "gdal>=3.0" from the https://github.com/insarlab/MintPy/blob/master/docs/conda.txt please and leave the issue open until basemap and GDAL3 are consistent (which shouldn't be long). This way we won't break mintpy (in particular basemap) for mintpy users. The solution @piyushrpt gave is clear enough for limited users of prep_aria.py at the moment.

yunjunz:
[...]

PR #131, opened 2019-07-23
installation: fix for basemap and pykml 

yunjunz:


yunjunz:
I have not added the note on basemap, because it seems like the development version of basemap has fixed this bug.

#83

yunjunz:
The development version of basemap works fine with me after testing. It fix the PROJ_LIB bug, thus, no manual setting in source file (.cshrc, .bash_prpfile) is needed anymore.

I removed `basemap` from `conda.txt` and `ports.txt` for now, to avoid running `uninstall` for fresh installation process; will add it back after basemap has a new release in macports and conda-forge.

PR #132, opened 2019-07-23
prep_aria: bug fix

ehavazli:
- minor bug fix in metadata
- add DEM option


yunjunz:
Hi @ehavazli is there any reason you closed this pull request?

ehavazli:
I found a minor bug in reading NoData value. Will re-open now

PR #134, opened 2019-07-23
plot_gps: search/use geometry file for incAngle if available

yunjunz:
+ utils.plot.plot_gps(): search existing geometry files with incidenceAngle dataset and use it to convert GPS enu to LOS.

+ utils.utils1.get_file_list(): bug fix related with coord argument

+ utils.utils1.get_geometry_file(): add work_dir to support searching while runing script not in the directory of the input file.

### DIFF
--- a/mintpy/objects/gps.py
+++ b/mintpy/objects/gps.py
@@ -234,35 +234,35 @@ class GPS:
                          + (self.std_u * unit_vec[2])**2)**0.5
         return self.dis_los, self.std_los
 
-    def get_los_geometry(self, insar_obj, print_msg=False):
+    def get_los_geometry(self, geom_obj, print_msg=False):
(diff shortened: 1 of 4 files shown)


ehavazli:
everything seems to be working except when I specify comparison to up component
```
--gps-comp up
```

yunjunz:
`--gps-comp up` is not really implemented, I removed this for now.

ehavazli:
consider changing the help output of view.py to remove up option so it won't be confusing.
Have you checked the codacy review?

yunjunz:
> consider changing the help output of view.py to remove up option so it won't be confusing.

Yes, it's removed from `utils.plot.add_gps_argument()` which is called by `view.py`. There is no extra `--gps-comp up` in view.py as far as I checked.

Thanks for the reminder of codacy suggestion.

PR #139, opened 2019-07-24
prep_aria: add more metadata

yunjunz:
prep_aria:

+ comment out skimage.transform.resize() because the updated ariaTSsteup.py produces the same size DEM file as the interferogram stack, this eliminates the need of installing scikit-image module while running prep_aria.py within the `ARIA-tools` python environment.

+ copy over all metadata from unwrappedPhase for the sake of completeness, except for Dates and bperp

+ add metadata extraction for:
   - ANTENNA_SIDE
   - HEADING
   - RANGE_PIXEL_SIZE
   - LON/LAT_REF1/2/3/4
   - CENTER_LINE_UTC

+ write all ifgrams in the sorted order of date1_date2 with date1 < date2

+ comment out geometry dataset azimuthAngle and shadowMask for now since both of them are not required by mintpy. Together with waterMask, they can be added afterward when the products are available from ARIA-tools.

+ more print out message for file IO

utils/ptime: remove obsolete ifgram_date_list() to that ptime.py can be independent from h5py module

add ARIA format link to README.md

Update indentation in smallbaselineApp.cfg

### DIFF
--- a/mintpy/prep_aria.py
+++ b/mintpy/prep_aria.py
@@ -1,15 +1,22 @@
 #!/usr/bin/env python3
+
+
 import os
-import gdal
-import h5py
+import time
 import argparse
+try:
+    import gdal
+except ImportError:
+    raise ImportError('gdal>=3.0 is required.')
+import h5py
 import numpy as np
+#from skimage.transform import resize
 from mintpy.utils import ptime
-from skimage.transform import resize
 
 
+####################################################################################
 EXAMPLE = """example:
-  prep_aria.py -w mintpy_SanFran -s stack/ -i incidenceAngle/20150605_20150512.vrt -d SRTM_3arcsec.dem
+  prep_aria.py -w mintpy -s stack/ -i incidenceAngle/20150605_20150512.vrt -d SRTM_3arcsec.dem
 """
 
 def create_parser():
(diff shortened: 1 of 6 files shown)


yunjunz:
Hi @ehavazli, could you check if all the operations you have used before work for you?

After this merge, I will cut a new release.

mintpy/prep_aria.py
```diff
@@ -1,216 +1,290 @@
...
     return
 
 
-def add_unwrapped_phase(h5File, unwStack, cohStack, connCompStack):
+def write_ifgram_stack(h5File, unwStack, cohStack, connCompStack):
 
+    print('-'*50)
+    print('opening {}, {}, {} with gdal ...'.format(os.path.basename(unwStack),
+                                                    os.path.basename(unwStack),
+                                                    os.path.basename(unwStack)))
```
ehavazli: all three lines are printing out unwStack

yunjunz: Just fixed.

mintpy/prep_aria.py
```diff
@@ -1,216 +1,290 @@
...
+    for key, value in metaUnw.items():
+        if key not in ["Dates", "perpendicularBaseline"]:
+            meta[key] = value
+
+    meta["ANTENNA_SIDE"] = -1
     meta["PROCESSOR"] = "isce"
     meta["FILE_LENGTH"] = ds.RasterYSize
     meta["LENGTH"] = ds.RasterYSize
-    meta["ORBIT_DIRECTION"] = "DESCENDING"
+    meta["ORBIT_DIRECTION"] = "DESCENDING"    #check with ARIA-tools team
```
yunjunz: Is there a way to extract this parameter, instead of hardwiring a fixed one? I have not looked into the stack vrt files in details yet.

ehavazli: I was looking into that and it is not in VRT's. I think it is good to add that information to the VRT files.

ehavazli:
everything ran without failing for a small test case (@hfattahi San Francisco data set) so it looks good to me.

mintpy/prep_aria.py
```diff
@@ -1,216 +1,290 @@
...
     meta["PLATFORM"] = "Sen"
-    meta["STARTING_RANGE"] = float(ds.GetRasterBand(1).GetMetadata("unwrappedPhase")['startRange'])
-    meta["WAVELENGTH"] = float(ds.GetRasterBand(1).GetMetadata("unwrappedPhase")['Wavelength (m)'])
+    meta["WAVELENGTH"] = float(meta["Wavelength (m)"])
     meta["WIDTH"] = ds.RasterXSize
     meta["NUMBER_OF_PAIRS"] = ds.RasterCount
 
-    # ARIA standard products currently don't number of range and
+    az_angle = float(meta["azimuthAngle"])
```
yunjunz: @ehavazli and @dbekaert, this value `azimuthAngle = -8.77992802893209` does not seem like the azimuthAngle of the satellite orbit (band2 of los.rdr from ISCE/topsStack); but more like the azimuth angle of the LOS vector (from incLocal.rdr), could you confirm?

If we got the HEADING angle of the satellite orbit, the `ORBIT_DIRECTION` can be drived from here.

ehavazli: azimuth angle in the VRT metadata is an average value. Azimuth angle information is extracted from the product. I am testing the new ariaTSsetup.py with the orbit direction information being included in the metadata of VRT.

mintpy/prep_aria.py
yunjunz: https://github.com/aria-tools/ARIA-tools/pull/75

yunjunz:
All metadata seems reasonable to me now. If one of two could approve this pull request, I will go forward and merge it.

yunjunz:
@ehavazli

PR #142, opened 2019-07-25
convert lookup-tables under geo-coord to radar-coord

ymcmrs:
Convert the Geo-coordinates based lookup-table (GAMMA, ROI_PAC) into the Radar-coordinates based lookup-table (ISCE).


yunjunz:
Could you check the suggestion by codacy?

PR #144, opened 2019-07-25
convert lookup-tables under geo-coord to radar-coord

ymcmrs:
Convert the Geo-coordinates based lookup-table (GAMMA, ROI_PAC) into the Radar-coordinates based lookup-table (ISCE).

### DIFF
--- a/mintpy/generate_rdc_lt.py
+++ b/mintpy/generate_rdc_lt.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python3
+############################################################
+# Program is part of MintPy                                #
+# Copyright(c) 2019    Yunmeng Cao                         #
+# Author:  Yunmeng Cao                                     #
+############################################################
+
+import os
+import sys
+import glob
+import time
+import argparse
+
+import h5py
+import numpy as np
+from mintpy.utils import readfile, writefile, ptime, utils as ut
+#########################################################################
+
+INTRODUCTION = '''
+#############################################################################
+   
+   Convert the Geo-coordinates based lookup-table (GAMMA, ROI_PAC) into the Radar-coordinates based lookup-table (ISCE).
+'''
+
+EXAMPLE = '''
+    Usage:
+            generate_rdc_lt.py geometryGeo.h5 
+            generate_rdc_lt.py geometryGeo.h5 -w geometryRadar.h5 
+            generate_rdc_lt.py geometryGeo.h5 -w geometryRadar.h5 -n 2
+
+##############################################################################
+'''
+def write_h5(datasetDict, out_file, metadata=None, ref_file=None, compression=None):
+    #output = 'variogramStack.h5'
+    'lags                  1 x N '
+    'semivariance          M x N '
+    'sills                 M x 1 '
+    'ranges                M x 1 '
+    'nuggets               M x 1 '
+    
+    if os.path.isfile(out_file):
+        print('delete exsited file: {}'.format(out_file))
+        os.remove(out_file)
+
+    print('create HDF5 file: {} with w mode'.format(out_file))
+    dt = h5py.special_dtype(vlen=np.dtype('float64'))
+
+    
+    with h5py.File(out_file, 'w') as f:
+        for dsName in datasetDict.keys():
+            data = datasetDict[dsName]
+            ds = f.create_dataset(dsName,
+                              data=data,
+                              compression=compression)
+        
+        for key, value in metadata.items():
+            f.attrs[key] = str(value)
+            #print(key + ': ' +  value)
+    print('finished writing to {}'.format(out_file))
+        
+    return out_file  
+
+def get_idx_cpx(RangeCoord,AzimuthCoord,WIDTH,LENGTH,n_lines):
+    
+    NN = n_lines # search n lines one time
+    x = np.arange(0,WIDTH)+1
+    y = np.arange(0,LENGTH)+1
+    xv, yv = np.meshgrid(x, y)
+    yv = yv.flatten()
+    xv = xv.flatten()
+    radar_cpx = xv + yv*1j
+    #print(len(radar_cpx))
+
+    
+    Ns = len(xv)
+    short_idx = np.zeros((Ns,),dtype=int)
+    
+    
+    gx = RangeCoord.flatten()
+    gy = AzimuthCoord.flatten()
+    gy[gy==0] = 10000
+    gy[gy<0] = 10000
+    geo_cpx = gx + gy*1j
+    
+    sort_gy_idx = np.argsort(gy)
+    sort_gy_array = gy[sort_gy_idx]
+    sort_gx_array = gx[sort_gy_idx]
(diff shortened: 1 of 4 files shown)


yunjunz:
Hi @ymcmrs, a self-explained script name is preferred, could you rename the new script to `lookup_geo2radar.py`, or your suggestion?

ymcmrs:
Done. 

As to gamma data, the loaded geomtryRadar.h5 does not have latitude and longitude, so it is necessary to generate lats/lons based on the four corners (e.g., for using pyaps under radar-coord), and I update this in the utilities0.py.  

ps: As to the older version of mintPy, the way (i.e., the function) to generate lats/lons under radar-coord is not right.

yunjunz:
Thank you @ymcmrs for the contributing!

Regarding the concerns about tropo_pyaps3.py for Gamma products in radar coordinates (same and accurate for products in geo-coordinates for all insar processors), I agree it's not an accurate calculation for the mesh-grid. From my limited testing, as shown below, I have not seen a significant difference yet. 

https://github.com/yunjunz/pyaps3/blob/master/examples/TestECMWF.ipynb

The ideal way would be providing the accurate lat/lon for each pixel, as your `lookup_geo2radar.py` would do.

Another way would be using the accurate "X/Y_FIRST" info instead of the inaccurate "LAT/LON_REF1/2/3/4" info (this info is not very accurate for roipac, how about Gamma?) to manually generate the mesh-grid.

yunjunz:
Hi @ymcmrs, regarding `lookup_geo2radar.py`, could you add more comments in the code about how you implemented it. I am clear about it yet.

From a simple test using the WellsEnvD2T399 example dataset, as shown below, the left side of both latitude and longitude does not seem reasonable.

geometryGeo

geometryRadar

ymcmrs:
Hi @yunjunz 
Thanks for the test. The theoretical precision of the lat/lon is the same as Y_STEP/X_STEP (assume the original geo-coord lookup table is right).
In your case, I doubt the original geo-coord lookup table is right? you can find the rangeCoord also looks not reasonable. 

In my case, LosAngeles area, the results looks fine.

Yunmeng

LosAngeles_geometryGeo
LosAngels_geometryRadar png

ymcmrs:
Here are the compared results between : 
(a) using geo-coord lookup table to convert the radar-coord lats/lons
(b) using the four-corners' lat/lon to convert the gridded radar-coord lats/lons

We can find over mountainous regions (also at the central part of the image), the gridded results have relatively large biases. (assume the geo-coord lookup based results are relatively better than gridded results), which is interesting, also reasonable I think.
Lat_Lon_diff_LT_Grid:lat
Lat_Lon_diff_LT_Grid:lon

ymcmrs:
@yunjunz  
The four-corners in ROI_PAC rsc means the SAR image? or the geocoded-SAR image?
If the corners mean that of the SAR corners (as GAMMA rsc), I am sure the way get_lat_lon is not right, but for the geo-SAR corners, it is no problem.

ymcmrs:
Here are the compared results between the generated lats/lons under radar coor-dinates:

old: using the current function get_lat_lon
new: using the corrected function get_lat_lon_rdc

lat_lon_new
lat_lon_old

ymcmrs:
Yes, I definitely know this.  I do not use a polynomial, and I also do not try to fit it, and what lookup_geo2rdc.py is to estimate the lat and lon pixel by pixel based on the available geo-coord lookup table, which considers the elevation during the generation in GAMMA. As to generation of the grid lasts/lons (when we do not have the lats or lons in geometryRadar), it is based on the assumption that increase of the lat/Lon for each pixel is the same, so when we know the four corners we can derive the lats lons matrixes.发自我的iPhone------------------ Original ------------------From: Zhang Yunjun  <notifications@github.com>Date: Fri,Jul 26,2019 6:08 PMTo: insarlab/MintPy <MintPy@noreply.github.com>Cc: Yunmeng Cao <ymcmrs@gmail.com>, Mention <mention@noreply.github.com>Subject: Re: [insarlab/MintPy] convert lookup-tables under geo-coord toradar-coord (#144)@yunjunz requested changes on this pull request.

Hi @ymcmrs, lookup table should take into account the topography and can not be determined by a two order polynomial. This effect is clear in WellsEnv case. This is the reason we use pyresample or scipy to do the interpolation instead of writing the code ourselves. Based on your test result, I would say it's tens-pixel level precision (always in practical precision). Since lookup_geo2radar.py is indepedent, I could still merge it and add comments about the precision so that one is aware of this issue while using it.
As a reference, you could check geocode.py --geo2radar option, which supports ISCE-style lookup table only, the conversion between geo and radar coordinates is very precise (sub-pixel level) as I tested. Currently, the support of --geo2radar option for gamma-style lookup table is missing (https://github.com/insarlab/MintPy/blob/master/mintpy/objects/resample.py#L360), I would recommend you to implement that if you are interested.
The current get_lat_lon() is only handling geo-coord input metadata, thus, very precise. Since this function is used in quite some other script, it should not be mixed with inaccurate solution. Could you remove the call of get_lat_lon_rdc() from it please?
For the purpose of running tropo_pyaps3.py for gamma radar-coord products, I would recommend calling get_lat_lon_rdc() directly within tropo_pyaps3.py when it is needed.

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/insarlab/MintPy/pull/144?email_source=notifications\u0026email_token=AFQZYFHOYRIZNFSXSCMPE6LQBMHPZA5CNFSM4IG4UB4KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOB7W5JPQ#pullrequestreview-267244734",
"url": "https://github.com/insarlab/MintPy/pull/144?email_source=notifications\u0026email_token=AFQZYFHOYRIZNFSXSCMPE6LQBMHPZA5CNFSM4IG4UB4KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOB7W5JPQ#pullrequestreview-267244734",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

ymcmrs:
Of course, this is only for GAMMA users( when try to use pyaps under radar coord), and ISCE users do not need to consider this, whose lookup table is under radar-coord, and geometryRadar contains lats and lons, and at the same time the corners in geometryGeo is like a rectangular.

――――――――――――――――
Yunmeng Cao, PhD,
Post Doctoral Fellow,
King Abdullah University of Science and
Technology (KAUST),
Crustal Deformation and InSAR (CDI) Group,
Jeddah, Saudi

________________________________

PR #146, opened 2019-07-25
bug fix for unwrap_erro_bridging if num_label > 10

yunjunz:
objects/conncomp.py:
+ fix bug of connection identification confusion when there are more than 10 reliable connected components in the dataset.

+ more comments on get_all_bridges() and find_mst_bridge()

save_roipac.py: support ouptut connectComponent from ifgramStack.h5 file

utils/readfile: more generic reading for binary data files by assigning some data structure info before checking InSAR processor

utils/writefile: bug fixed for writing byte and bool type of data file

specify ECMWF (ERAI) as the tropo dataset for test example. They should be changed to ERA5 in the future.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,7 +7,7 @@
 [![License](https://img.shields.io/badge/license-GPL-yellow.svg)](https://github.com/insarlab/MintPy/blob/master/LICENSE)
 [![Forum](https://img.shields.io/badge/forum-Google%20Group-orange.svg)](https://groups.google.com/forum/#!forum/mintpy)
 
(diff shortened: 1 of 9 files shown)


yunjunz:
@ehavazli from my testing, with either ifgramStack.h5 file or with single ISCE interferogram (*.unw and *.unw.conncomp file), it works. But I don't have a dataset with large connected components more than 10.

hfattahi:
@yunjunz 
I don't know what is the exact problem that you guys want to solve. But keep in mind that :
1) MintPy currently enforces np.byte datatype for connected components dataset. This means only values between -128 to 127 can be stored. 
2) ISCE2 snaphu wrapper, outputs maximum 20 connected component and zeros out other components by default. But the largest can be up to 255 in Snaphu as the data type is unsigned byte (needs to be verified). Therefore np.byte in MintPy is fine for ISCE regular outputs.
3) arai tools stitches different unwrapped interferograms and relables the connected components. So it is possible to have large number of connected components over large areas.
4) Other unwrappers (currently outside ISCE2) allow for much larger number of connected components

With all this I'm wondering if we should change the data type of connected compinent dataset to float !

hfattahi:
You are fast. While I was writing you merged :). I'll open an issue then.

yunjunz:
;)

Issue #148, opened 2019-07-25
data type for connected components dataset


hfattahi:
1) MintPy currently enforces np.byte datatype for connected components dataset. This means only values between -128 to 127 can be stored.
2) ISCE2 snaphu wrapper, outputs maximum 20 connected component and zeros out other components by default (can be controlled). But the largest can be up to 255 in Snaphu as the data type is unsigned byte (needs to be verified). Therefore np.byte in MintPy is fine for ISCE regular outputs.
3) arai tools stitches different unwrapped interferograms and relabels the connected components. So it is possible to have large number of connected components over large areas.
4) Other unwrappers (currently outside ISCE2) allow for much larger number of connected components

With all this I'm wondering if we should change the data type of connected components in MintPy hdf5 dataset to float?

hfattahi:
@dbekaert @ehavazli What is the largest number of connected components that you guys may expect?

dbekaert:
20 per product. So 20*n products people stitch.
I know of people stitching 8-9 products, so like 180  connected components or so.

yunjunz:
the `connectComponent` object in mintpy is calling [skimage.measure.label()](https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.label) to extract all regions/areas related properties, and this `label` function supports integer array as input only, thus, we could change the data type from `byte` to [`int16`](https://docs.scipy.org/doc/numpy-1.13.0/user/basics.types.html), which supports up to 32767. How about that?

hfattahi:
That is more than enough. The only implication is that we will increase the size on disk, by moving from np.byte to anything else. If increasing size is not a big concern for users then I think we should move to int16 .

yunjunz:
Since the value of this dataset is usually small, compression should reduce the size a lot. We could turn ON the compression for this dataset always.

hfattahi:
Yes I agree. So let's move to int16 then.

PR #150, opened 2019-07-26
 connectComponent: byte to int16

yunjunz:
prep_aria.py and objects/stackDict.py:

+ change the data type of connectComponent dataset from BYTE to INT16 to support up to 32767 reliable regions/connected components for single interferograms.

+ turn ON compression of "lzf" while writing connectComponent to HDF5 file to save disk space

prep_aria.py:
+ add FILE_TYPE metadata
+ add MODIFICATION_TIME metadata to 3D datasets
+ turn ON chunks while writing HDF5 to speedup file IO
+ turn ON maxshape for 3D datasets to support updatable ifgramStack.h5: feature not used anyway yet; just to be consistent with mintpy.

update default plot settings for connectComponent dataset in view.py and  plot.py


yunjunz:
#148

PR #155, opened 2019-07-26
prep_aria: no-data for waterMask, unw, coh and connComp

yunjunz:
Grab mask of invalid pixels from incidence angle and use it to maskout pixels in water mask too. This benefits:

1) more representative calculation of the average spatial coherence --> improving the performance of coherence-based network modification; 

2) one way of handling the no-data because water mask is used as a basic mask in mintpy.

### DIFF
--- a/mintpy/prep_aria.py
+++ b/mintpy/prep_aria.py
@@ -245,9 +245,22 @@ def write_ifgram_stack(h5File, unwStack, cohStack, connCompStack):
     dsCoh = gdal.Open(cohStack, gdal.GA_ReadOnly)
     dsComp = gdal.Open(connCompStack, gdal.GA_ReadOnly)
 
-    nPairs = dsUnw.RasterCount
+    # extract NoDataValue (from the 2nd */date2_date1.vrt file for example)
+    ds = gdal.Open(dsUnw.GetFileList()[2], gdal.GA_ReadOnly)
+    noDataValueUnw = ds.GetRasterBand(1).GetNoDataValue()
+    print('grab NoDataValue for unwrapPhase:      {:<5} and convert to 0.'.format(noDataValueUnw))
+
+    ds = gdal.Open(dsCoh.GetFileList()[2], gdal.GA_ReadOnly)
+    noDataValueCoh = ds.GetRasterBand(1).GetNoDataValue()
+    print('grab NoDataValue for coherence:        {:<5} and convert to 0.'.format(noDataValueCoh))
+
+    ds = gdal.Open(dsComp.GetFileList()[2], gdal.GA_ReadOnly)
+    noDataValueComp = ds.GetRasterBand(1).GetNoDataValue()
+    print('grab NoDataValue for connectComponent: {:<5} and convert to 0.'.format(noDataValueComp))
+    ds = None
 
     # sort the order of interferograms based on date1_date2 with date1 < date2
+    nPairs = dsUnw.RasterCount
     d12BandDict = {}
     for ii in range(nPairs):
         bnd = dsUnw.GetRasterBand(ii+1)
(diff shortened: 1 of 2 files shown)


yunjunz:
#100

hfattahi:
@yunjunz It's strange. I only see a small part of your commit which is just reading the waterMask in prep_aria.py. Can you check if everything is fine?

hfattahi:
never mind looks like you already merged part of it!

yunjunz:
I should have waited for merging!

mintpy/prep_aria.py
```diff
@@ -267,22 +280,26 @@ def write_ifgram_stack(h5File, unwStack, cohStack, connCompStack):
 
         h5["date"][ii,0] = d12.split("_")[0].encode("utf-8")
         h5["date"][ii,1] = d12.split("_")[1].encode("utf-8")
+        h5["dropIfgram"][ii] = True
 
         bnd = dsUnw.GetRasterBand(bndIdx)
-        h5["unwrapPhase"][ii,:,:] = -1.0*bnd.ReadAsArray()
+        data = bnd.ReadAsArray()
+        data[data == noDataValueUnw] = 0      #assign pixel with no-data to 0
+        h5["unwrapPhase"][ii,:,:] = -1.0*data #date2_date1 -> date1_date2
```
dbekaert: @yunjunz just double checking if we are in sync.

For ARIA standard product:
Reference (master) is acquired later in time than secondary (slave). 
Product sign convention: land uplift corresponds to a range decrease (negative radians).

yunjunz:
Hi @hfattahi, the new commit grab the no-data value using gdal function for uwn, coh and connComp dataset, instead of using hardwired -1 for connComp only; and convert them to zero. This won't change any loading result so far, but should be more robust for the future.

dbekaert:
@yunjunz this is great!

Quick question, does Mintpy only skips the time-series estimation of pixels based on the mask being provided or does it also check based on the 0 value in the IFGs?
e.g. At some point you mentioned Nan versus 0.

yunjunz:
@dbekaert, mintpy skip all of the three criteria you mentioned: skip time-series estimation if any of them meets.

mintpy/prep_aria.py
hfattahi: @yunjunz can confirm better.
In Mintpy outputs, uplift is positive and subsidence is negative (consistent with ARIA). The -1 here is because of the date swap. Based on our offline discussions, ARIA products always have date2_date1 form while Mntpy expects the opposite. The datelist datset is sorted as MintPy expects so we put -1 here.

yunjunz: MintPy convention: land uplift (moving towards satellite) corresponds to negative phase value in radian. It seems the same...

Could you check your aria+mintpy result on Mauna Loa (Sen desc track 87 IW2) with the one below? It should be inflating on the right side (black cross in the map).
https://insarmaps.miami.edu/start/19.7/-155.6/7.5?onlyShowDatasets=S1_IW2_087_0528_0532_20141116_XXXXXXXX,&zoomOut=false

Screen Shot 2019-07-26 at 10 07 47 PM

hfattahi: @yunjunz from the figure you posted, uplift (right side red signal) is positive. Isn't it?

yunjunz: @hfattahi yes, it is positive as phase (displacement) in meters; and mintpy expects an equivalent interferogram with negative unwrapped phase in radian. At the end of invert_network step, mintpy multiply all phase_in_radian by ( wavelength / (-4pi)).

hfattahi: That's right and I was referring to MintPy outputs.

hfattahi: The sign of the final products in the San Fran dataset makes sense with MintPy convention. If @dbekaert confirms the sign in the final MintPy product of Hawaii dataset, then we should be good to merge this PR.

dbekaert: I concur same sign for Hawaii.  I made it a while ago after @hfattahi made a change on the sign per Brett's request to make it work  for Norfolk. I do not think it has been changed since then.

yunjunz: I guess that’s enough for confirm the sign is right.

PR #161, opened 2019-07-31
check azimuthAngle while plotting gps vs insar

yunjunz:
As for now, GPS enu2los conversion for ARIA products would work if the azimuth angle is not loaded and not written in geometryGeo.h5 file, then the code will use incidence angle and head angle calculated from metadata. For a small area, this would be fine.

The azimuthAngle inconsistency from ARIA-tools and topsStack is not exactly 90 degrees difference, thus, I don't have a solution until Emre or David give me a clear explanation #162.


yunjunz:
As for now, GPS enu2los conversion for ARIA products would work if the azimuth angle is not loaded and written in geometryGeo.h5 file, then the code will use incidence angle and head angle calculated from metadata. For a small area, this would be fine.

yunjunz:
#162

PR #169, opened 2019-08-02
 add docs/conda_env.yml for mintpy environment installation

yunjunz:


yunjunz:
https://github.com/aria-tools/ARIA-tools/issues/71

Issue #172, opened 2019-08-07
Timeseries from ARIA standard products


rzinke:
Hello,

I am trying to run the MintPy timeseries for a very basic test case data set for ARIA standard product data over Tibet (ascending track 041).
 
I went through the setup steps required in the ARIA tools time series setup workflow:
ariaDownload.py -t 041 -b ’37.5 40 87 91’ -s 20180601 -e 20181001
ariaTSsetup.py -f .products/’S1*.nc’ -d Download -nt 8
python3 prep_aria.py -s stack -d DEM/SRTM_3arcsec.dem.vrt -i incidenceAngle/20180613_20180601.vrt
smallbaselineApp.py –start modify_network
 
The MintPy processes goes through the first several steps (modify network, reference point, correct unwrap error [off], stack interferograms, invert network, correct lod [off], troposphere [off], deramp). Then, it gets 22% of the way through the “correct topography” step before throwing an error:
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/smallbaselineApp.py", line 1070, in <module>
    main()
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/smallbaselineApp.py", line 1060, in main
    app.run(steps=inps.runSteps, plot=inps.plot)
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/smallbaselineApp.py", line 1016, in run
    self.run_topographic_residual_correction(sname)
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/smallbaselineApp.py", line 796, in run_topographic_residual_correction
    mintpy.dem_error.main(scp_args.split())
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/dem_error.py", line 489, in main
    inps = correct_dem_error(inps, A_def)
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/dem_error.py", line 425, in correct_dem_error
    num_step=num_step)
  File "/u/sarh0/rzinke/tools/MintPy/mintpy/dem_error.py", line 308, in estimate_dem_error
    X = linalg.lstsq(A, ts, cond=1e-15)[0]
  File "/u/sar-r0/rzinke/python/miniconda3/envs/MintPy/lib/python3.7/site-packages/scipy/linalg/basic.py", line 1152, in lstsq
    a1 = _asarray_validated(a, check_finite=check_finite)
[...]

PR #176, opened 2019-08-07
 fix bug of mask from geometry in dem_error.py

yunjunz:
+ dem_error.py: ignore pixels with NaN value in any of the geometry dataset, which caused the initial error in #172.

+ prep_aria.py: add more notes in example

+ smallbaselineApp.py: move waterMask.h5 generation from step `load_data` to `modify_network`, so that `smallbaselineApp.py --dostep load_data` is equivalent to `load_data.py` and `prep_aria.py`.

### DIFF
--- a/mintpy/defaults/smallbaselineApp.cfg
+++ b/mintpy/defaults/smallbaselineApp.cfg
@@ -34,7 +34,7 @@ mintpy.subset.yx   = auto    #[1800:2000,700:800 / no], auto for no
 mintpy.subset.lalo = auto    #[31.5:32.5,130.5:131.0 / no], auto for no
 
 
-## 2 Modify Network (optional)
+## 2 Modify Network
 ## 1) Coherence-based network modification = Threshold + MST, by default
 ## It calculates a average coherence value for each interferogram using spatial coherence and input mask (with AOI)
 ## Then it finds a minimum spanning tree (MST) network with inverse of average coherence as weight (keepMinSpanTree)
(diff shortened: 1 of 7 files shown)


yunjunz:
My concern is: waterMask.h5 is generated **before** modify_network, while avgSpatialCoh.h5 and maskConnComp.h5 are generated after the optional **modify_network**. Since waterMask.h5, if available, is used in several steps, any step that includes this, load_data or modify_network, can not be optional. Therefore, I actually prefer the old way: writing waterMask.h5 within `load_data` step in smallbaselineApp.py. 

`load_data.py` and `prep_aria.py` are equivalent to each other; `smallbaselineApp.py --dostep load_data` was not equivalent to them before this commit. Even with the native isce+mintpy workflow, one should not run `load_data.py` and `smallbaselineApp.py --start modify_network`.

After mintpy is able to generate water mask in geo and radar coordinates, for all insar processors, including isce and gamma, we should add a new step `generate_water_mask` between `load_data` and `modify_network` and put it as a required step.

Both `avgSpatialCoh.h5` and `maskConnComp.h5` are used only once afterwards, therefore, they can be generated right before where they are used, out of the `modify_network` step. 

With all above, `modify_network` then can be optional. And no extra step, i.e. `generate_aux_file` is needed. What do you think @hfattahi ?

hfattahi:
It's really a minor issue. Let's get back to this later.

yunjunz:
Sounds good. Let's get back to this later.

PR #179, opened 2019-08-09
shrink image size with ImgBot

ehavazli:
5 figures are optimized and reduced in size  by ImgBot.


yunjunz:
lossless compression by ImgBot.

PR #212, opened 2019-09-09
Bug fixes for unwrapping error correction using bridging method for network modified ifgramStack file

ehavazli:
- In case of more than 10 connected components, previous versions sorting method failed (previously line 213). Added a comparison between connected component numbering.

- When labelRef is 0, previously, line 203 throws an overflow error due to negative value as starting point. Added an if clause and passed 0 as starting point in case labelRef is 0. Please check if this would affect the rest of the process.


yunjunz:
> When labelRef is 0, previously, line 203 throws an overflow error due to negative value as starting point. Added an if clause and passed 0 as starting point in case labelRef is 0. Please check if this would affect the rest of the process.

labelRef is the label value of the reference reliable region, it should not be zero, because "zero" regions are the area that are not selected by connected components. Since this variable is assigned at here: https://github.com/insarlab/MintPy/blob/master/mintpy/objects/conncomp.py#L72. The zero labelRef value occurred because in this interferogram, the reference point is not in the connected component. 

The current script is applying bridging to all interferograms, but it should not apply to dropped interferograms. I will modify the unwrap_error_briding.py accordingly.

yunjunz:
Hi @ehavazli, could you check if the commits above work for you?

This error happens very occasionally and I can not test on my dataset.

ehavazli:
hey @yunjunz,

yes the code works for me with the changes you made.

yunjunz:
Great! I will go forward and merge the fix.

PR #242, opened 2019-11-18
Save time-series in QGIS compatible format

piyushrpt:

Adds support to export MintPy time-series in a QGIS ps-time-series plugin compatible shapefile format - https://plugins.qgis.org/plugins/pstimeseries/ . Thanks @yunjunz for help with figuring out the directory structure. 

Usage:
```
  save_qgis.py timeseries_ERA5_ramp_demErr.h5 -g inputs/geometryRadar.h5 -o ts.shp
  save_qgis.py geo/geo_timeseries_ERA5_ramp_demErr.h5 -g geo/geo_geometryRadar.h5 -o ts.shp
```

TO DO: Possibly export to pgsql for faster load speeds than shape files.  Will be done in separate PR after some experimentation.

**Reminders**

### DIFF
--- a/mintpy/save_qgis.py
+++ b/mintpy/save_qgis.py
@@ -69,7 +69,7 @@ def gatherMintpy(ts_file, geom_file):
             }
 
     #Check if the files exists.
-    for key, fname in rdict.items():
+    for fname in rdict.values():
         if not os.path.isfile(fname):
             raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), fname)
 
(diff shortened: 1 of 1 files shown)


falkamelung:
Does this (or will it) support Cloud-Optimized Format? It would be really cool to have those files sitting on an S3 or wrangler storage and look at them without the need to download.

piyushrpt:
@falkamelung This might be able to load remote datasets. Performance might not be that impressive though. Performance on a local machine with few million points seems to be ok and this allows one to use tools like profiles etc that are available as well supported plugins.

Issue #164, opened 2019-08-01
Confusing, non-fatal `basemap` installation errors under linux


falkamelung:
I am getting the building errors below when installing MintPy as described in installallation.md. It looks very scary as the errors show up red (not  reproduced here) but  they are not fatal.  It says `failed to install basemap` and a bit later `Successfully installed basemap-1.2.0`. Very confusing!   Can't we go back to `conda install basemap` which did not throw any errors  (I forgot why this had to be changed). If not we should add a note that these errors are nothing to worry about.

 Here what I did. I install as :
```
./Miniconda3-4.5.12-Linux-x86_64.sh -b -p ./miniconda3
./miniconda3/bin/conda install --yes --file MintPy/docs/conda.txt 
./miniconda3/bin/pip install git+https://github.com/matplotlib/basemap.git#egg=mpl_toolkits
./miniconda3/bin/pip install git+https://github.com/tylere/pykml.git
```
I am getting the following errors, but at the end it says 'Sucessfully':
```
[famelung@pegasus test_mintpy]$ ./miniconda3/bin/pip install git+https://github.com/matplotlib/basemap.git#egg=mpl_toolkits
Collecting mpl_toolkits from git+https://github.com/matplotlib/basemap.git#egg=mpl_toolkits
  Cloning https://github.com/matplotlib/basemap.git to /tmp/pip-install-ag112_58/mpl-toolkits
  Running command git clone -q https://github.com/matplotlib/basemap.git /tmp/pip-install-ag112_58/mpl-toolkits
  WARNING: Generating metadata for package mpl-toolkits produced metadata for project name basemap. Fix your #egg=mpl-toolkits fragments.
Requirement already satisfied: matplotlib!=3.0.1,>=1.0.0 in ./miniconda3/lib/python3.7/site-packages (from basemap) (3.1.1)
Requirement already satisfied: numpy>=1.2.1 in ./miniconda3/lib/python3.7/site-packages (from basemap) (1.16.4)
Requirement already satisfied: pyproj>=1.9.3 in ./miniconda3/lib/python3.7/site-packages (from basemap) (2.2.1)
Collecting pyshp>=1.2.0 (from basemap)
Requirement already satisfied: six in ./miniconda3/lib/python3.7/site-packages (from basemap) (1.12.0)
[...]

PR #262, opened 2020-01-10
Added geos and basemap as conda packages to conda_env.yml

rzinke:
Modified conda_env.yml file to include geos and basemap modules as conda installations; removed pip install basemap. This is intended to resolve issues with pip finding the proper geos lib files when installing basemap.


yunjunz:
#164

PR #294, opened 2020-03-10
GPS stations not plotted if coincident with mask

sssangha:

I have updated “view.py” and “utils/plot.py” such that GPS stations which spatially coincide with masked areas—and/or nodata values if the ‘zero mask’ option is specified—are now also excluded from the visualization (see attached image below ).

Screen Shot 2020-03-10 at 12 17 08 AM

This is done by indexing the mask array for the pixel corresponding to a given GPS station. If a spatial subset/crop is specified during visualization, the mask array and corresponding metadata object are appropriately updated under “utils/plot.py”

### DIFF
--- a/mintpy/view.py
+++ b/mintpy/view.py
@@ -143,8 +143,8 @@ def cmd_line_parse(iargs=None):
         inps.save_fig = True
     if inps.lalo_step:
         inps.lalo_label = True
-    if inps.zero_mask:
-        inps.mask_file = 'no'
+    #!#if inps.zero_mask:
+        #!#inps.mask_file = 'no'
 
     if not inps.disp_whitespace:
         inps.disp_axis = False
@@ -956,12 +956,18 @@ def read_data4figure(i_start, i_end, inps, metadata):
     # mask
     if inps.msk is not None:
         vprint('masking data')
-        msk = np.tile(inps.msk, (data.shape[0], 1, 1))
-        data = np.ma.masked_where(msk == 0., data)
+        inps.msk = np.tile(inps.msk, (data.shape[0], 1, 1))
+        data = np.ma.masked_where(inps.msk == 0., data)
+
+    if inps.msk is None:
+        inps.msk = np.tile(np.ones(data.shape, dtype=np.int8), (data.shape[0], 1, 1))
 
     if inps.zero_mask:
         vprint('masking pixels with zero value')
-        data = np.ma.masked_where(data == 0., data)
+        inps.msk = np.ma.masked_array(inps.msk, mask=np.isnan(data))
+        np.ma.set_fill_value(inps.msk, 0.)
+        inps.msk=inps.msk.filled()
+
     return data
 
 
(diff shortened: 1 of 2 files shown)


yunjunz:
Hi @sssangha, thank you for contributing! I think this is a nice feature.

I can see you are trying to track the mask matrix, which was missing in the code because I am not very clear about numpy masked array and np.nan, both of which are plotted as transparent in matplotlib. Another complexity is the 2D and 3D matrix: the input data can be 2D or 3D, but we want to keep the mask matrix as 2D only for the index operation, such as you have here. Thus, I do not recommend you to track the mask matrix unless you are VERY comfortable with view.py.

Since `plot_gps` is only supported in the one_plot scenarios here [view.py#L498](https://github.com/insarlab/MintPy/blob/master/mintpy/view.py#L498), I would suggest adding another argument, i.e. mask, to plot_gps() and grabbing the mask array from the final 2D data matrix right before calling [plot_gps()](https://github.com/insarlab/MintPy/blob/master/mintpy/utils/plot.py#L1234) without dealing with mask during the whole process.

mintpy/view.py
```diff
@@ -956,12 +956,18 @@ def read_data4figure(i_start, i_end, inps, metadata):
     # mask
     if inps.msk is not None:
         vprint('masking data')
-        msk = np.tile(inps.msk, (data.shape[0], 1, 1))
-        data = np.ma.masked_where(msk == 0., data)
+        inps.msk = np.tile(inps.msk, (data.shape[0], 1, 1))
+        data = np.ma.masked_where(inps.msk == 0., data)
+
+    if inps.msk is None:
```
yunjunz: This is the 3D input data scenario, where we don't want to save the 3D mask matrix.

mintpy/view.py
```diff
@@ -956,12 +956,18 @@ def read_data4figure(i_start, i_end, inps, metadata):
...
+        inps.msk = np.tile(np.ones(data.shape, dtype=np.int8), (data.shape[0], 1, 1))
 
     if inps.zero_mask:
         vprint('masking pixels with zero value')
-        data = np.ma.masked_where(data == 0., data)
+        inps.msk = np.ma.masked_array(inps.msk, mask=np.isnan(data))
```
yunjunz: The previous data masking operation line is missing here.

mintpy/view.py
```diff
@@ -1342,12 +1353,17 @@ def plot(self):
...
                 vprint('masking data')
                 data = np.ma.masked_where(self.msk == 0., data)
+            if self.msk is None:
+                self.msk = np.ones(data.shape, dtype=np.int8)
+            if self.zero_mask:
+                vprint('masking pixels with zero value')
+                self.msk = np.ma.masked_array(self.msk, mask=np.isnan(data))
+                np.ma.set_fill_value(self.msk, 0)
+                self.msk=self.msk.filled()
+                data = np.ma.masked_where(self.msk == 0., data)
```
yunjunz: This modification does not work as expected, for the Fernandina example dataset: the water body, which has zero velocity, should be masked out.

```
view.py velocity.h5 velocity --mask no --zero-mask
```

Screen Shot 2020-03-10 at 3 17 56 PM

PR #300, opened 2020-03-16
Read list of dates in textfile for mintpy.network.referenceFile

bbuzz31:
It seems to me that MintPy should be able to read a text file of interferograms to include in the network (e.g. `mintpy.network.referenceFile = date12_list.txt`).
However, the function call seems to only read hdf5 files.
I wrote up a quick workaround:

The call to read the `mintpy.network.referenceFile` is wrapped in a try/except; it will proceed as usual if given a path to hdf5 file. 

otherwise, it calls the function I wrote `get_date12_txt`, which reads the file as text and searches each line for two dates (ref, secondary) of form YYYYmmdd. The regex i used should handle any separator (and any number of them) between the two dates. The order of the two dates per line does not matter.

Please let me know if this functionality already exists somewhere else, or what additional steps I can take to get this code in the way you would like. Thanks!

### DIFF
--- a/mintpy/modify_network.py
+++ b/mintpy/modify_network.py
@@ -312,7 +312,7 @@ def get_date12_to_drop(inps):
 
     # reference file
     if inps.referenceFile:
-        date12_to_keep = ifgramStack(inps.referenceFile).get_date12_list(dropIfgram=True)
+        date12_to_keep = pnet.get_date12_list(inps.referenceFile, dropIfgram=True)
         print('--------------------------------------------------')
         print('use reference pairs info from file: {}'.format(inps.referenceFile))
         print('number of interferograms in reference: {}'.format(len(date12_to_keep)))
--- a/mintpy/objects/stack.py
+++ b/mintpy/objects/stack.py
@@ -837,37 +837,14 @@ class ifgramStack:
 
     # Functions considering dropIfgram value
     def get_date12_list(self, dropIfgram=True):
-        try:
-            with h5py.File(self.file, 'r') as f:
-                dates = f['date'][:]
-                if dropIfgram:
-                    dates = dates[f['dropIfgram'][:], :]
-            mDates = np.array([i.decode('utf8') for i in dates[:, 0]])
-            sDates = np.array([i.decode('utf8') for i in dates[:, 1]])
-        except:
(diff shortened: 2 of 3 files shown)


yunjunz:
Hi @bbuzz31, thank you for pointing out this issue and fix. 

Yes, there is a function get_date12_list() in utils/network.py. You could add the following to [L182](https://github.com/insarlab/MintPy/blob/master/mintpy/utils/network.py#L182) to convert the date string format into YYYYMMDD_YYYYMMDD style:

```
date12_list = ptime.yyyymmdd_date12(date12_list)
```

A cleaner way to add support to smallbaselineApp.py would be: in modify_network.py, replacing [L315](https://github.com/insarlab/MintPy/blob/master/mintpy/modify_network.py#L315)

```
date12_to_keep = ifgramStack(inps.referenceFile).get_date12_list(dropIfgram=True)
```

with 

```
date12_to_keep = pnet.get_date12_list(inps.referenceFile, dropIfgram=True)
```

Could you please commit this when you got a chance?

bbuzz31:
Hi @yunjunz thanks for the suggestion and pointing to the internal function.
I replaced [L315 ](https://github.com/insarlab/MintPy/blob/master/mintpy/modify_network.py#L315) in modify_network.py as you suggested.
However, there were a couple of issues with `pnet.get_date12_list`
The first issue was taking the first interferogram only of the textfile utils/network ([L177](https://github.com/insarlab/MintPy/blob/master/mintpy/utils/network.py#L177))
Then I had to reduce the dimensionality (to correctly) convert to a list. 
The list format is needed for the set difference that comes in `modify_network.py`, [L319](https://github.com/insarlab/MintPy/blob/master/mintpy/modify_network.py#L319)

I did a recursive grep on the mintpy directory, and it doesn't look like there should be any conflicts. I think the only time this function is called it is passed an hdf5 array, so the text reader doesn't get called.

yunjunz:
Hi @bbuzz31, thank you for the commit. You are right, this function is not really used in the current code. I just fixed the list format issue. Let me know if this works for you please, then I will go forward to merge this request.

bbuzz31:
It works, thanks @yunjunz!

yunjunz:
Great!

PR #263, opened 2020-01-31
Replace Basemap with Cartopy

Ovec8hkin:
**Description of proposed changes**
Replaced `Basemap` library with `cartopy` library for future support. 

Matplotlib has opted to end support for geographic plotting using Basemap and has switcher to using `cartopy` for mapping ([See More](https://matplotlib.org/basemap/users/intro.html#cartopy-new-management-and-eol-announcement)). Newer versions of `matplotlib` are expected to completely deprecate the use of Basemap. 

**Reminders**

- [x] Fix #164, #171
- [x] Pass Codacy code review (green)
- [x] Pass testing with `$MINTPY_HOME/test/test_smallbaselineApp.py`
- [x] Make sure that your code follows our style. Use the other functions/files as a basis.
- [x] If modifying functionality, describe changes to function behavior and arguments in a comment below the function declaration.

PR #303, opened 2020-03-21
Cartopy final

Ovec8hkin:
Final carroty updates for scalebar sizing

**Reminders**

- [ ] Fix #xxxx


yunjunz:
Hi @Ovec8hkin, this pull request is outdated because in #263 I found that the `CartopyScalebar` object is not needed, as we could adjust the old draw_scalebar() to use `pyproj` only and be independent from basemap and cartopy.

Ovec8hkin:
Ok. Will remove.

Issue #36, opened 2019-04-14
minimum version of disk_jobqueue that contains LSFCluster


yunjunz:
Hi @2gotgrossman, could you add a minimum version constarint in `docs/conda.txt` file for `dask_jobqueue`, both I and @mirzaees encountered the issue that conda install an obsolete version of 0.1, which does not have LSFCluster. I fixed the issue by re-installed a new miniconda3, which later on install the dask_jobqueue with version 0.4.

https://github.com/insarlab/PySAR/blob/d3cac2b0d4c4e97b01a5fb637c5382f43bcf023d/docs/conda.txt#L4

mirzaees:
@2gotgrossman @yunjunz 
I installed the dask_jobqueue=0.4 but then I get following error which askes for dask distributed. I follow the instruction but nothing changes:


2019-04-15 09:46:31,186 - INFO - pysarApp.py /nethome/sxm1611/insarlab/infiles/sxm1611/TEMPLATES/test_unittestGalapagosSenDT128.template --end load_data |& tee out_pysar.log
20190415:094631 * pysarApp.py /nethome/sxm1611/insarlab/infiles/sxm1611/TEMPLATES/test_unittestGalapagosSenDT128.template --end load_data |& tee out_pysar.log
Traceback (most recent call last):
  File "/nethome/sxm1611/test/testsara/rsmas_insar/3rdparty/miniconda3/lib/python3.6/site-packages/dask/distributed.py", line 5, in <module>
    from distributed import *
  File "/nethome/sxm1611/test/testsara/rsmas_insar/3rdparty/miniconda3/lib/python3.6/site-packages/distributed/__init__.py", line 5, in <module>
    from .actor import Actor, ActorFuture
  File "/nethome/sxm1611/test/testsara/rsmas_insar/3rdparty/miniconda3/lib/python3.6/site-packages/distributed/actor.py", line 4, in <module>
    from .client import Future, default_client
  File "/nethome/sxm1611/test/testsara/rsmas_insar/3rdparty/miniconda3/lib/python3.6/site-packages/distributed/client.py", line 29, in <module>
    from dask.optimization import SubgraphCallable
ImportError: cannot import name 'SubgraphCallable'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/nethome/sxm1611/test/testsara/rsmas_insar/sources/PySAR/pysar/pysarApp.py", line 21, in <module>
[...]

PR #307, opened 2020-03-22
remove Dask version constraint for installation?

Ovec8hkin:
Removed version dependencies for dask and dask-jobqueue. Thorough testing on both Pegasus and Stampede have shown that the version dependencies are unnecessary and may cause issues.

**Reminders**

### DIFF
--- a/docs/conda.txt
+++ b/docs/conda.txt
@@ -2,8 +2,8 @@ python>=3.6
 cartopy
 cdsapi
 cvxopt
-dask>=1.0,<2.0
-dask-jobqueue>=0.3,<1.0
+dask
+dask-jobqueue
 ecCodes
 h5py
 lxml
--- a/docs/conda.txt
+++ b/docs/conda.txt
@@ -2,8 +2,8 @@ python>=3.6
 cartopy
 cdsapi
 cvxopt
-dask
-dask-jobqueue
+dask>=1.0
+dask-jobqueue>=0.3
 ecCodes
 h5py
 lxml


yunjunz:
From the previous experience #36, adding a minimum version requirement solved some installation issues. Thus, I recommend keeping the min version number.

Generally, I am fine with removing the max version number constraint. I am unclear why @2gotgrossman add it initially.

Ovec8hkin:
Ok. We can probably retain the minimum version numbers, though Im not sure why anyone would install old enough version of the dask modules to create an issue.

2gotgrossman:
Some projects introduce breaking changes when the major version number changes. I think adding the max version # was an attempt to prevent breaking changes occurring due to newer releases. https://en.wikipedia.org/wiki/Software_versioning#Degree_of_compatibility

I do wonder why this issue occurred...My only thought is that a Dask dependency was updated and a newer version of Dask handles that change. If this were the case, the only way around this issue would be to freeze all dependencies with a certain version #.

yunjunz:
The latest `Dask` version is [2.12](https://docs.dask.org/en/latest/changelog.html) and the latest `dask-jobqueue` version is [0.7](https://jobqueue.dask.org/en/latest/changelog.html). As long as the latest version is supported, I feel fine to remove the max version number. What do you think @2gotgrossman? Could you confirm this @Ovec8hkin?

Same rule as above for the min version number, if a fresh installation (using conda for example), without the min version number constraint, works fine. It would be nice to remove this too. Could you please also confirm this? @Ovec8hkin

Ovec8hkin:
@yunjunz Both Falk and I are able to run `ifgram_inversion.py` on Stampede and Pegasus using `dask 2.12.0`, `dask-core 2.12.0`, and `dask-jobqueue 0.7.0` so I think we can remove the minimum version numbers.

yunjunz:
Sounds good to me.

Ovec8hkin:
@2gotgrossman I can not comment about why using the maximum version numbers no longer work. I do know we need them removed now, because we have to take advantage of the current master of `dask-jobqueue` rather than the current version available through conda/pip (they have a fix to a bug we are suffering from in master that isn't available through conda/pip yet).

Ovec8hkin:
@yunjunz I am going to leave the minimum version numbers in there for now. If everything is working fine after a few more installations I will remove it .

PR #315, opened 2020-04-05
support dask configuration by config_name

Ovec8hkin:
Added support for specifying a dask configuration (specified in the `dask.yaml` file) to use by its `config_name`. If none is provided, the default configuration name (lowercased version of the cluster type) is used.

**Reminders**


mintpy/ifgram_inversion.py
```diff
@@ -1084,7 +1086,10 @@ def ifgram_inversion(ifgram_file='ifgramStack.h5', inps=None):
         # at once (other user's jobs gained higher priority in the general at that point)
         NUM_WORKERS = inps.numWorker
         # FA: the following command starts the jobs
-        cluster = cl.get_cluster(type=inps.cluster, walltime=inps.walltime)
+        if inps.clusterConfig == "None":
```
yunjunz, line 1089: I recommend changing the default value to `no` because: `None` from default parser in ifgram_inversion.py and `"None"` from template file are not consistent with each other, there should be a bug here.

Ovec8hkin: Yes. I was wondering how to do handle this case. I didn't like having string None, but I wasn't sure how else to deal with it. "no" makes sense as a default value.

yunjunz:
Hi @Ovec8hkin, I did the following changes, for simplicity. Do you see issues with them?

1. rename `mintpy.networkInversion.clusterConfig` to `mintpy.networkInversion.config`.
2. In the installation note, rename `~/.config/dask/dask.yaml` to `~/.config/dask/mintpy.yaml`
3. move the checking of --config option to the cmd_line_parse().

mintpy/ifgram_inversion.py
```diff
@@ -196,6 +198,21 @@ def cmd_line_parse(iargs=None):
...
+    # --config option
+    inps.config = inps.config.lower()    # use lowercase all the time
+    if inps.config == 'no':
+        inps.config = ipns.cluster.lower()
+
+    if inps.parallel and inps.config != inps.cluster.lower():
+        import dask
+        try:
+            dask.config.get('jobqueue.{}'.format(inps.config))
+        except KeyError:
```
yunjunz, line 210: If the custom input dask configuration not found. Is it better to `raise ValueError()` and exit the program, or to just printout a warning message and continue?

I usually prefer to raise exception and exit, so that the user can be more aware of the type errors and avoid using unnecessary computing time.

Ovec8hkin: Since there is a sensible fallback option, I would rather use the fallback than violently terminate the program. I figure the printed message should be sufficient.

mintpy/ifgram_inversion.py
```diff
@@ -196,6 +198,21 @@ def cmd_line_parse(iargs=None):
 
     inps.timeseriesFile, inps.tempCohFile = inps.outfile
 
+    # --config option
+    inps.config = inps.config.lower()    # use lowercase all the time
+    if inps.config == 'no':
```
Ovec8hkin, line 203: Wouldn't it be better to move this condition to after the config auto values have been sourced so we can just say `if inps.config` rather than having to do this string comparison?

mintpy/defaults/smallbaselineApp.cfg
```diff
@@ -119,9 +119,10 @@ mintpy.networkInversion.minRedundancy   = auto #[1-inf], auto for 1.0, min num_i
 
 ## Parallel processing with Dask for HPC
 mintpy.networkInversion.parallel    = auto #[yes / no], auto for no, parallel processing using dask
-mintpy.networkInversion.cluster	    = auto #[SLURM / PBS / LSF], auto for SLURM, cluster type for dask
-mintpy.networkInversion.numWorker   = auto #[int > 0], auto for 40, number of works for dask cluster to use
-mintpy.networkInversion.walltime    = auto #[HH:MM], auto for 00:40, walltime for dask workers
+mintpy.networkInversion.cluster     = auto #[SLURM / PBS / LSF], auto for SLURM, cluster type
+mintpy.networkInversion.config      = auto #[SLURM / PBS / LSF / no], auto for no (same as cluster), configuration name
+mintpy.networkInversion.numWorker   = auto #[int > 0], auto for 40, number of works to deploy
+mintpy.networkInversion.walltime    = auto #[HH:MM], auto for 00:40, walltime for each dask worker
```
Ovec8hkin, line 125: We need to be a bit careful here, because LSF takes wall time in HH:MM format, but SLURM and PBS take it in HH:MM:SS format. I would have to think about the best way to handle this.

mintpy/ifgram_inversion.py
```diff
@@ -1070,6 +1088,7 @@ def ifgram_inversion(ifgram_file='ifgramStack.h5', inps=None):
     # Parallel loop
     else:
         try:
+            import dask
```
Ovec8hkin, line 1091: If you move the config_name fallback logic to `create_parser` this import statement can be removed.

mintpy/ifgram_inversion.py
```diff
@@ -209,12 +226,13 @@ def read_template2inps(template_file, inps):
     keyList = [i for i in list(iDict.keys()) if key_prefix+i in template.keys()]
     for key in keyList:
         value = template[key_prefix+key]
+        print(key, value)
```
Ovec8hkin, line 229: This can also be removed.

mintpy/ifgram_inversion.py
yunjunz: The auto value from smallbaselineApp.cfg file is replaced within read_template2inps(), so yes, this is after that. I changed the default value for --config option in this script to "no" so that both default value from ifgram_inversion.py and smallbaselineApp.cfg file are the same. Then, we translate the config=='no' here to a more meaningful value.

mintpy/defaults/smallbaselineApp.cfg
yunjunz: `mintpy.objects.cluster.get_cluster()` sounds like a good option.

mintpy/ifgram_inversion.py
yunjunz: Agree.

mintpy/ifgram_inversion.py
yunjunz: We need the fallback logic after read_template2inps().

mintpy/defaults/smallbaselineApp.cfg
Ovec8hkin: Yes. That is probably the best place for it. I guess I'll just check the length of string and the cluster type and remove/append the ':SS' part as needed. 

So if its `LSF` but gives an 8 character wall time, I can remove the last 3 characters to make it 'HH:MM' format, or if its `SLURM` but gives a 5 character wall time, I can append ':00' to the end to force 'HH:MM:SS' format. Thoughts on this strategy?

mintpy/ifgram_inversion.py
Ovec8hkin: This import can still be removed, because `dask` will have already been imported before this code every runs.

mintpy/ifgram_inversion.py
Ovec8hkin: The string comparison feels unclean to me. I don't see it anywhere else in the codebase, even for parameters that use 'no' as default values.

mintpy/defaults/smallbaselineApp.cfg
yunjunz: Your way sounds good to me.

mintpy/ifgram_inversion.py
yunjunz: I did not realize it, you are right.

mintpy/ifgram_inversion.py
yunjunz: As long as it's clear to users and works, it's fine with me. We can commit this and come back to it later when we have a cleaner solution.

yunjunz: `inps.config = inps.config.lower()` is to ensure it works for scenarios when one entered upper case config. Just a robust checking.

It's a little bit of confusion in smallbaselineApp.cfg file that we use upper case for cluster and lower case for config for the default values. Thus, I adjust the comments in file to all UPPER case except for "no", to be more consistent.

Ovec8hkin: I was actually going to fix this so that the cluster type was capitalization-agnostic.

yunjunz: That would be even better!

mintpy/ifgram_inversion.py
yunjunz: Now I remembered. `dask` has to be imported here because the one imported in cmd_line_parse() is a local variable that won't be existed within the ifgram_inversion().

yunjunz:
I revoked your last commits, because:
1. we still need to translate `mintpy.networkInversion.config = no`.
2. dask import in cmd_line_parse() should as less as possible, more conditions checking helps.
3. the variable `dask` imported in cmd_line_parse() is local and will be lost in ifgram_inversion(), thus, need to be imported again. 
4. job_queue is still imported in objects.cluster, thus should be included in the error message.

yunjunz:
It looks all good to me now. @Ovec8hkin Could you please do a final test and confirm it works as expected? Then we could squash and merge it (write some final commit message for easy tracking and reference please).

mintpy/objects/cluster.py
```diff
@@ -31,18 +46,37 @@ def get_cluster(cluster_type, **kwargs):
     return cluster
 
 
-def format_walltime_for_cluster(cluster, wall):
-    wall_len = len(wall)
-    formatted_wall = wall
-    if cluster in ['pbs', 'slurm'] and wall_len != 8:
-        if wall_len == 5:
-            formatted_wall = wall + ":00"
+def check_config_name(config_name, cluster_type)
```
Ovec8hkin, line 49: I would expect any function that starts with the name `check_` to return a boolean indicating whether or not the inputs pass the check, but not to deal with any formatting of the inputs to force obedience to the check. I would rename these functions to `format_{param}`

mintpy/objects/cluster.py
```diff
@@ -31,18 +46,37 @@ def get_cluster(cluster_type, **kwargs):
...
+    config_names = list(dask.config.get('jobqueue').keys())
+    if config_name not in config_names:
+        msg = 'Dask configuration "{}" was not found in ~/.config/dask/*.yaml'.format(config_name)
+        msg += '\nFall back to default config name: "{}"'.format(cluster_type)
+        config_name = cluster_type
+
+    return config_name
+
+
+def check_walltime_format(walltime, cluster_type):
```
Ovec8hkin, line 64: Same thing here.

mintpy/objects/cluster.py
yunjunz: Your suggestion sounds good to me. Go for it please.

Ovec8hkin:
Added ability for users to specify a dask configuration name in the `smallbaselineApp.cfg` file for easy switching between dask configurations based on system or style.

Also included parsing to handle switching between HH:MM and HH:MM:SS wall time styles for various schedulers.

mintpy/objects/cluster.py
yunjunz: This is now done in #351.

PR #319, opened 2020-04-12
support `aria` processor in smallbaselineApp.py 

yunjunz:

This PR support `mintpy.load.processor = aria` in smallbaselineApp.py, so that users do not need to manually run prep_aria.py anymore. Instead, one just need to setup the template file and run smallbaselineApp.py directly after ARIA-tools commands.

An example directory structure with template options for ARIA is provided in the doc.

An example template file for the San Francisco dataset is uploaded.

The jupyter notebook example in  need to be updated as well, it will be in another PR in MintPy-tutorial repo.

**Reminders**

- [x] Pass testing with `$MINTPY_HOME/test/test_smallbaselineApp.py`


dbekaert:
@yunjunz that is great, and makes the mintpy experience complete.

PR #328, opened 2020-04-25
add scikit-learn to the dependency for bootstrapping

ehavazli:
- prep_aria.py metadata read module has been fixed.
- new dependency required for boot_strap.py added to docs.

**Reminders**

- [ ] Fix #xxxx


dbekaert:
@ehavazli @yunjunz this might have been an inconsistency introduced in ARIA-tools during our latest set of edits to add support for S3. If that is the case we should fix in ARIA-tools, if not we fix in prep_aria. both code bases have been changes.

@yunjunz Can you confirm no changes in mintpy on prep_aria reading step for the timing?

dbekaert:
Looks like it was introduced in ARIA-tools. We fixed it on our end.
@ehavazli could you back out the time fix?

yunjunz:
Good to hear it's cleared out.

PR #329, opened 2020-04-26
bugfix in prep_aria

sssangha:
Since my latest PR to ARIA-tools ensures consistency with the decimal seconds convention, I reverted the changes made here to capture errors associated with passing time in absolute seconds.

Also fixed bug address in ARIA-tools issue 147 (aria-tools/ARIA-tools#147). Specifically, prep_aria crashes when one IFG passed (relevant for unit tests).


sssangha:
@yunjunz ok, I went ahead and squashed the conflicts.

PR #345, opened 2020-05-12
plot_smallbaselineApp.sh: running view.py commands in parallel

falkamelung:
Add an '&' to all view.py commands and a 'wait' after the last one, so that they all run in parallel.  On stampede this reduces the runtime from 2 minutes to 10 seconds.  On systems that have less cores than commands  it is supposed to run as much commands in parallel as cores are available. This worked for me on linux (16 core machine, pegasus login node). It should behave equally on a Mac, but I did not try.
**Reminders**

- [x] Pass testing with `$MINTPY_HOME/test/test_smallbaselineApp.py`

### DIFF
--- a/sh/plot_smallbaselineApp.sh
+++ b/sh/plot_smallbaselineApp.sh
@@ -45,82 +45,83 @@ view='view.py --nodisplay --dpi 150 --update '
 opt=' --dem '$dem_file' --mask '$mask_file' -u cm '
 #opt=' --dem '$dem_file' --mask '$mask_file' -u cm --vlim -2 2'
 if [ $plot_key_files -eq 1 ]; then
-    file=velocity.h5;              test -f $file && $view $file $opt               >> $log_file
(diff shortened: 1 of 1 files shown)


yunjunz:
Hi @falkamelung, I recently realized that this change introduced a big increase of memory usage and led to failure for big datasets. I will have to revert this change unless you have a better way to handling the memory issue.

falkamelung:
Hmmm.... We will have the conflict between powerful and not powerful systems frequently. It would be could to have a solution. There are several other functions that could be run in parallel. 

We could have a mintpy.compute.plotParallel option . 

On powerful systems dask will used. If mintpy.cluster is on, it could be switched on. Otherwise it is off.

I am not so happy that the compute options ended up inside smallbaselineApp.cfg but I did not want to argue.  We should thing about a compute.cfg or compute.yaml next to smallbaselineApp.cfg (only used if it exists).  But this is for another time.

yunjunz:
The switching on and off sounds good to me. If you or others contribute, I am more than happy to merge it.

yunjunz:
I will go ahead to revert this then.

PR #386, opened 2020-06-26
defaults/auto_path: added reading of ISCE watermask

falkamelung:
Reads ISCE_produced watermask

**Reminders**


yunjunz:
@falkamelung As far as I know, only stripmapStack produce waterMask.rdr. So do you want to remove the default path for topsStack?

falkamelung:
I did not know. I see. It does not raise an exception if there is no watermask file so we can leave it.  It  should be easy to implement in stripmapStack.  Mintpy does not create a watermask, or does it?

yunjunz:
Sounds good to me, as long as it does not give an error. MintPy does not create water mask.

PR #389, opened 2020-06-28
utils1.get_lookup_file(): revert work_dir arg to fix bug

yunjunz:

revert the previouly introduced work_dir argument to fix the bug it introduced; and replace it by adding more directory candicates to achive the same functionality.

**Reminders**

- [x] Fix a bug introduced in #388

### DIFF
--- a/mintpy/objects/coord.py
+++ b/mintpy/objects/coord.py
@@ -45,9 +45,7 @@ class coordinate:
         """
         self.src_metadata = metadata
         if lookup_file is None:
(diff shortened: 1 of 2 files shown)


yunjunz:
This PR pass the $MINTPY_HOME/test/test_smallbaselineApp.py on my laptop, however, it did not pass the same test on circle CI. While the previous PR #388 passed the test on circle CI but did not pass the test on my laptop. @mehdizadehm Could you please take a look to check why this happens?

PR #403, opened 2020-07-18
update default path to the new ISCE diversity naming style

mirzaees:

Changed the auto default path based on new changes in ISCE which has renamed a few files and folders

**Reminders**

### DIFF
--- a/mintpy/smallbaselineApp.py
+++ b/mintpy/smallbaselineApp.py
@@ -208,6 +208,8 @@ class TimeSeriesAnalysis:
     def __init__(self, customTemplateFile=None, workDir=None):
         self.customTemplateFile = customTemplateFile
         self.workDir = os.path.abspath(workDir)
+        if not 'mintpy' in self.workDir:
+            self.workDir = self.workDir + '/mintpy'
         self.cwd = os.path.abspath(os.getcwd())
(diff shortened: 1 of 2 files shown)


mintpy/smallbaselineApp.py
```diff
@@ -208,6 +208,8 @@ class TimeSeriesAnalysis:
     def __init__(self, customTemplateFile=None, workDir=None):
         self.customTemplateFile = customTemplateFile
         self.workDir = os.path.abspath(workDir)
+        if not 'mintpy' in self.workDir:
+            self.workDir = self.workDir + '/mintpy'
```
yunjunz, line 212: To have `mintpy` in the working directory, one can user `--dir` option to specify that manually. This change would force the user to have `mintpy` in their directory path. I thought you would not like it, considering combing with minopy workflow. Are you sure about this?

mirzaees: If it is not there, it produces error while I am in project directory. but you are right with --dir there is no problem. I leave it to you to decide
No problem with minopy

yunjunz: Let's keep the old way then, I think it's more consistent on the user side.

PR #436, opened 2020-09-05
Improve memory efficiency in the steps of calculating ramp, deramp, residual_RMS

liuykai:

Deramp the dataset day-by-day instead of doing it at once with a large array.
Calculate the time-series residual RMS of the dataset day-by-day instead of doing it at once with a large array.

When dealing with large datasets (large width, length with many dates in time series), we may have a memory issue when trying to allocate a big array. So we chop it down to day-by-day calculation.

In addition, a syntax bug in readfile.py is also spotted and fixed. 

**Reminders**

### DIFF
--- a/mintpy/utils/utils1.py
+++ b/mintpy/utils/utils1.py
@@ -716,12 +716,42 @@ def run_deramp(fname, ramp_type, mask_file=None, out_file=None, datasetName=None
 
     # deramping
     if k == 'timeseries':
-        print('reading data ...')
-        data = readfile.read(fname)[0]
+        # initialize dataset structure
+        ts_obj     = timeseries(fname)
+        ts_obj.open()
+        date_list  = ts_obj.dateList
+        num_date   = len(date_list)
+        length     = int(atr['LENGTH'])
+        width      = int(atr['WIDTH'])
+        date_dtype = np.dtype('S{}'.format(len(date_list[0])))
+        dsNameDict = {
+            "date"       : (date_dtype, (num_date,)),
+            "bperp"      : (np.float32, (num_date,)),
+            "timeseries" : (np.float32, (num_date, length, width)),
+        }
+     
+        # write HDF5 file with defined metadata and (empty) dataset structure
+        writefile.layout_hdf5(out_file, dsNameDict, atr)
+
+        # write date time-series
+        date_list_utf8 = [dt.encode('utf-8') for dt in date_list]
+        writefile.write_hdf5_block(out_file, date_list_utf8, datasetName='date')
+
+        # write bperp time-series
+        bperp = ts_obj.pbase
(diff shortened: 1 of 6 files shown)


mintpy/utils/utils1.py
```diff
@@ -719,7 +719,8 @@ def run_deramp(fname, ramp_type, mask_file=None, out_file=None, datasetName=None
         print('reading data ...')
         data = readfile.read(fname)[0]
         print('estimating phase ramp ...')
-        data = deramp(data, mask, ramp_type=ramp_type, metadata=atr)[0]
+        for i in range(data.shape[0]):
+            data[i,:] = deramp(data[i,:], mask, ramp_type=ramp_type, metadata=atr)[0]
```
hfattahi, line 723: The entire dataset seems to have been still loaded in to memory. It would be nice to load only what is needed to memory (which is one date at a time) do the computation, write that date and then go to the next. Otherwise this wouldn't be still memory friendly even for multi-looked data.

hfattahi (commented):
This has been an issue for a while and its really great to see you are trying to fix it. Thanks @liuykai . As i left a comment, it would be great to load only one date at a time to memory whenever possible. 
One more thing to note is that, in case of very big datasets, I don't really think if even the entire data for a date is needed to compute ramps. One could only load a sample of the data to compute ramp e.g., instead of using every pixel, maybe every other N pixel (N to be determined) would be enough. I can imagine that N could default to 1 (i.e., all pixels) and be set for higher numbers for large datasets. I understand if this will be addressed in a future PR.

mintpy/utils/utils1.py
liuykai: I see. Thanks for the comment, and I will try to modify it.

yunjunz:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- mintpy/utils/utils1.py  1
- mintpy/ifgram_inversion.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/insarlab/MintPy/pullRequest?prid=6131547&bid=20053259)

yunjunz:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- mintpy/utils/utils1.py  1
- mintpy/ifgram_inversion.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/insarlab/MintPy/pullRequest?prid=6131547&bid=20053259)

yunjunz (approved):
This is great! Thank you @liuykai!

yunjunz:
The coarse sampling while reading has been implemented recently via `readfile.read(xstep, ystep)`, it should be easily added in the future in another PR I assume.

liuykai:
@yunjunz and @hfattahi, thanks for your guidance.

PR #468, opened 2020-11-06
add setup.py for pip install and pypi distribution  

yunjunz:

+ add setup.py for pip install and pypi distribution with a [wiki note](https://github.com/insarlab/MintPy/wiki/pip-install)

**Reminders**


yunjunz:
There are two things missing:
+ environment variable `$MINTPY_HOME` is not set; this is optional but would make a lot of documentation easy.
+ the `$MINTPY_HOME/mintpy` path is not added into `$PATH`, thus making the scripts there not available in the command line if I run `pip install --index-url https://test.pypi.org/simple/ --no-deps mintpy_test4==1.2.3` to install.

Any idea how to do these @davidwgrossman @dbekaert @piyushrpt?

PR #483, opened 2020-11-30
bring back tropo_gacos.py for geo/radar coord

mirzaees:

+ refactor tropo_gacos.py to be consistent with the latest mintpy version with block-by-block IO

+ support `mintpy.troposphericDelay.method = gacos` in smallbaselineApp.py

+ fix a bug in diff.py on reference point/date when splitting into boxes

+ fix a bug in subset.subset_dataset() when input is binary file

**Reminders**


mintpy/tropo_gacos.py
```diff
@@ -0,0 +1,366 @@
...
+            print('No reference info found in input file, use input ref_yx: ' + str(inps.ref_yx))
+            atr['REF_Y'] = inps.ref_yx[0]
+            atr['REF_X'] = inps.ref_yx[1]
+    
+    # ****reading incidence angle file***/
+    if os.path.isfile(inps.inc_angle):
+        inps.inc_angle = readfile.read(inps.inc_angle, datasetName='incidenceAngle')[0]
+        inps.inc_angle = np.nan_to_num(inps.inc_angle)
+    else:
+        inps.inps.inc_angle = float(inps.inc_angle)
```
yunjunz, line 211: Typo here?

mirzaees: yes thank you!

mintpy/objects/stack.py
```diff
@@ -269,7 +269,7 @@ def read(self, datasetName=None, box=None, squeeze=True, print_msg=True):
                 box = [0, 0, self.width, self.length]
 
             # read
-            data = ds[dateFlag,
+            data = ds[np.ix_(dateFlag)[0],
```
yunjunz, line 272: From my testing, the index issue is related to h5py v3+. I am using Python 3.8 with h5py-2.10, and it seems working fine for me. 

Assuming the h5py v3+ bug will be fixed in the future, I think it's better if we just wait and stick to version 2+ for now. #474 adds version constrain to h5py for the installation. Could you confirm if it works for you?

mirzaees: I see, I have installed h5py separately and that is probably the issue. So shall I undo this one? or you will do it?

yunjunz: If you could do it, please.

mintpy/defaults/smallbaselineApp.cfg
```diff
@@ -183,6 +183,11 @@ mintpy.troposphericDelay.polyOrder      = auto  #[1 / 2 / 3], auto for 1
 mintpy.troposphericDelay.looks          = auto  #[1-inf], auto for 8, extra multilooking num
 mintpy.troposphericDelay.minCorrelation = auto  #[0.0-1.0], auto for 0
 
+## Notes for gacos:
+## GACOS products should be downloaded manually before running this step.
+## They have to be one .ztd binary file for each acquisition date,
+## and the path to these downloaded products should be given here:
+mintpy.troposphericDelay.gacosDir      = auto # [path2directory], auto for GACOS_DIR or "./"
```
yunjunz, line 190: Since the GACOS products are named as (YY)YYMMDD.ztd format, it can not really be shared among different projects. I would recommend not use ${WEATHER_DIR}, but set it to `./GACOS` directly, inside the mintpy folder.

Could you please also add the GACOS website link and the relevant paper citation here so that user could easily refer to?

mintpy/smallbaselineApp.py
```diff
@@ -769,6 +772,19 @@ def get_dataset_size(fname):
...
+                import glob
+                tropo_file = self.workDir + '/GACOS.h5'
+                GACOS_dir = self.template['mintpy.troposphericDelay.gacosDir']
+                iargs = [in_file, '-l', geom_file, '-i', geom_file, '-o', tropo_file, '--GACOS-dir', GACOS_dir]
+                print('tropospheric delay correction with gacos approach')
+                print('\ntropo_gacos.py', ' '.join(iargs))
+                if ut.run_or_skip(out_file=out_file, in_file=in_file) == 'run':
+                    mintpy.tropo_gacos.main(iargs)
+                    correct_timeseries(in_file, tropo_file, out_file)
+
```
yunjunz, line 786: The `mintpy.tropo_pyaps3.correct_timeseries()` is a very short function, I would recommend to copy the code from tropo_pyaps3.py into tropo_gacos.py so that 1) the import here can be simplified and 2) without import tropo_pyaps3.py, which checks pyaps3.

yunjunz (changes requested):
Thank you for your effort to bring back GACOS @mirzaees, this is definitely desired by a lot of users.

The previous framework of tropo_gacos.py is pretty old, I would recommend a few big changes:
1. make timeseries_file and geom_file (which should always have the required geometry datasets) as required inputs, to simplify the geometry and metadata preparation in the main().
2. add a copy of correct_timeseries() from tropo_pyaps3.py to here, so that the output of this script could be `./inputs/GACOS.h5` (in the same coord as the ifgramStack.h5, in geo- or radar- coordinates) and `timeseries_GACOS.h5`
3. Drop the support of non-isce processors for now to simplify the code of get_delay(), since nobody is using roipac anymore.
4. For isce(aria) products in geo-coordinates, use `skimage.transform.resize()` as shown in the [GACOS notebook](https://nbviewer.jupyter.org/github/insarlab/MintPy-tutorial/blob/main/applications/tropo_gacos.ipynb), it's much faster and way simpler than RGI.
5. for ISCE products in radar-coordinates, pass the interpolated/resampled delay matrix directly, without writing to a .npy file, either by saving the data of all dates into a big 3D matrix, or by using the block-by-block IO via writefile.layout_hdf5() and writefile.wirte_hdf5_block() as shown in https://github.com/yunjunz/MintPy/blob/main/mintpy/tropo_pyaps3.py#L677.

Let me know what do you think please.

mintpy/smallbaselineApp.py
yunjunz: For `mintpy.tropo_gacos.main()` to work, you might need to add it work the `workflow/__init__.py`.

mirzaees:
Hi @yunjunz, I agree with the proposed changes, please check and let me know if more is required that I am missing

yunjunz (approved):
It looks all good to me. Thank you @mirzaees!

I tested time-series file in geo / radar coord, the script runs without issue. Could you confirm and maybe also test one example using .unw file?

mirzaees:
@yunjunz, I test all now and it works

Issue #493, opened 2020-12-27
timeseries2velocity list index out of range


RussellGrew:
Thanks for making this great tool available open source.

**Description of the problem**

I have prepared my first ISCE Sentinel stack using topsStack.py and I am trying to run MintPy for the first time also. The smallbaselineApp fails during the velocity calculation.

**Note that I only have three dates in the stack**. I am assuming this is sufficient to test the workflow end to end. 

I based the configuration on the example https://mintpy.readthedocs.io/en/stable/dir_structure/

**Full script that generated the error**

The full output of smallbaselineApp.py is below, including the error

```bash
-RUN-at-2020-12-27 18:26:45.259118--
Current directory: /media/storage/isce_stack
Run routine processing with smallbaselineApp.py on steps: ['load_data', 'modify_network', 'reference_point', 'quick_overview', 'correct_unwrap_error', 'invert_network', 'correct_LOD', 'correct_troposphere', 'deramp', 'correct_topography', 'residual_RMS', 'reference_date', 'velocity', 'geocode', 'google_earth', 'hdfeos5']
Remaining steps: ['modify_network', 'reference_point', 'quick_overview', 'correct_unwrap_error', 'invert_network', 'correct_LOD', 'correct_troposphere', 'deramp', 'correct_topography', 'residual_RMS', 'reference_date', 'velocity', 'geocode', 'google_earth', 'hdfeos5']
--------------------------------------------------
Go to work directory: /media/storage/isce_stack
read default template file: /media/storage/isce_stack/smallbaselineApp.cfg


******************** step - load_data ********************

load_data.py --template /media/storage/isce_stack/smallbaselineApp.cfg
processor : isce
SAR platform/sensor : unknown from project name "None"
--------------------------------------------------
prepare metadata files for isce products
prep_isce.py -m /media/storage/isce_stack/reference/IW1.xml -g /media/storage/isce_stack/merged/geom_reference -b /media/storage/isce_stack/baselines -d /media/storage/isce_stack/merged/interferograms -f filt_*.unw
['/media/storage/isce_stack/reference/data.rsc'] exists and is newer than ['/media/storage/isce_stack/reference/IW1.xml'] --> skip.
prepare .rsc file for geometry files
read perp baseline time-series from /media/storage/isce_stack/baselines
prepare .rsc file for  filt_*.unw
[==================================================] 20201209_20201221    0s /     0s
Done.
--------------------------------------------------
searching interferometric pairs info
input data files:
[...]

PR #494, opened 2020-12-28
tsview.py to handle case of three data points in the stack

RussellGrew:

Details are here https://github.com/insarlab/MintPy/issues/493

tsview.py was failing when only three InSAR images in the stack


yunjunz (approved):
Thanks, @RussellGrew.

PR #519, opened 2021-02-16
gacos: support new .tif format

sssangha:
For GACOS-corrections, subsets were not properly applied. Specifically, the entire GACOS field was passed. Thus, this would result in inconsistent dimension errors:

```
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/smallbaselineApp.py", line 1182, in <module>
    main(sys.argv[1:])
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/smallbaselineApp.py", line 1172, in main
    app.run(steps=inps.runSteps, plot=inps.plot)
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/smallbaselineApp.py", line 1122, in run
    self.run_tropospheric_delay_correction(sname)
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/smallbaselineApp.py", line 823, in run_tropospheric_delay_correction
    mintpy.tropo_gacos.main(iargs)
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/tropo_gacos.py", line 340, in main
    calculate_delay_timeseries(tropo_file=inps.tropo_file,
  File "/u/sar-r2/ssangha/conda_installation/stable_aug14_2020/MintPy/mintpy/tropo_gacos.py", line 192, in calculate_delay_timeseries
    ztd_files = np.array(ztd_files)[flag].tolist()
IndexError: boolean index did not match indexed array along dimension 0; dimension is 202 but corresponding boolean dimension is 101
```

After implementing my fixes, subsets were taken as expected:

Screen Shot 2021-02-16 at 11 59 15 AM

Furthermore, GACOS products appear to be distributed as ‘TIF’ files now by default, while Mintpy had still expected products with the original ‘.ztd’ extension. Now a check for both has been integrated.

### DIFF
--- a/docs/references.md
+++ b/docs/references.md
@@ -36,6 +36,10 @@
 
 + Werner, C., U. Wegmüller, T. Strozzi, and A. Wiesmann (2000), Gamma SAR and interferometric processing software, paper presented at _Proceedings of the ERS-Envisat symposium_, Gothenburg, Sweden.
 
++ Yu, C., Z. Li, and N. T. Penna (2018), Interferometric synthetic aperture radar atmospheric correction using a GPS-based iterative tropospheric decomposition model, Remote Sensing of Environment, 204, 109-121, doi:10.1016/j.rse.2017.10.038.
+
++ Yu, C., Z. Li, N. T. Penna, and P. Crippa (2018), Generic Atmospheric Correction Model for Interferometric Synthetic Aperture Radar Observations, Journal of Geophysical Research: Solid Earth, 123(10), 9202-9222, doi:10.1029/2017JB015305.
+
 + Yunjun, Z., H. Fattahi, and F. Amelung (2019), Small baseline InSAR time series analysis: Unwrapping error correction and noise reduction, _Computers & Geosciences, 133_, 104331, doi:10.1016/j.cageo.2019.104331.
 
 + Zebker, H. A., P. A. Rosen, and S. Hensley (1997), Atmospheric effects in interferometric synthetic aperture radar surface deformation and topographic maps, _Journal of Geophysical Research: Solid Earth, 102_(B4), 7547-7563, doi:10.1029/96JB03804.
--- a/mintpy/defaults/smallbaselineApp.cfg
+++ b/mintpy/defaults/smallbaselineApp.cfg
@@ -178,8 +178,8 @@ mintpy.solidEarthTides = auto #[yes / no], auto for no
 ##      ERA5  - ERA-5       from ECMWF [need to install PyAPS from GitHub; recommended and turn ON by default]
 ##      MERRA - MERRA-2     from NASA  [need to install PyAPS from Caltech/EarthDef]
 ##      NARR  - NARR        from NOAA  [need to install PyAPS from Caltech/EarthDef; recommended for N America]
-## c. gacos - use GACOS with the iterative tropospheric decomposition model (Yu et al., 2017, JGR; 2018, JGR; 2018, RSE)
-##      Manually download GACOS products at http://www.gacos.net/index.html for all acquisitions before running this step.
+## c. gacos - use GACOS with the iterative tropospheric decomposition model (Yu et al., 2018a, RSE; 2018b, JGR)
+##      Manually download GACOS products at http://www.gacos.net for all acquisitions before running this step.
 mintpy.troposphericDelay.method = auto  #[pyaps / height_correlation / gacos / no], auto for pyaps
 
 ## Notes for pyaps:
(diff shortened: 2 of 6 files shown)


sssangha:
@yunjunz as an update to this PR, I’ve also captured a few additional bugs in the `reference_date.py` script. Specifically the following:

1. For certain products that didn’t have a predefined reference date (e.g. cubes containing the absolute delays), the following bug was raised so this condition must be adjusted accordingly:
```
  File "/u/tiger-z0/ssangha/conda_installation/stable_oct15_2020/MintPy/mintpy/reference_date.py", line 226, in <module>
    main(sys.argv[1:])
  File "/u/tiger-z0/ssangha/conda_installation/stable_oct15_2020/MintPy/mintpy/reference_date.py", line 209, in main
    change_timeseries_ref_date(ts_file,
  File "/u/tiger-z0/ssangha/conda_installation/stable_oct15_2020/MintPy/mintpy/reference_date.py", line 129, in change_timeseries_ref_date
    if ref_date == atr['REF_DATE']:
```

2. The metadata of the input file was updated, instead of that of the output file (if different).

3. All of the output layers were empty:
Screen Shot 2021-02-16 at 5 14 52 PM

This was fixed by emulating the manner in which files were written out in the `reference_point.py` script:
Screen Shot 2021-02-16 at 5 15 11 PM

yunjunz:
Thank you @sssangha for addressing this issue. Is it possible for you to send me a test dataset (timeseries file, geometry file and the downloaded gacos files)?

yunjunz:
Thank you for identifying all 3 bugs in `reference_date.py` @sssangha! The zero value in the output file is due to an incorrect index in the `block` variable, introduced at #488. I keep the old way of calling `writefile.write_hdf5_block()` for memory usage reason.

For other users who might be interested/concerned about their results from `smallbaselineApp.py`. With these bugs, the script changes the `REF_DATE` value and not update the matrix value. It does not affect the velocity estimation or displaying results from tsview.py, which re-referencing on the fly always; it does give an inconsistent zero-value acquisition date when plotting using view.py.

yunjunz:
I sneaked in a few docs update on the reference, hope you don't mind ;)

yunjunz:
Hi @sssangha, thank you for sending me the test data. I guess the subset issue might be because you used `tropo_gacos.py -f geo_timeseries.he -g input/geometryRadar.h5`, instead of `tropo_gacos.py -f geo_timeseries.h5 -g geo_geometryRadar.h5`?

For `tropo_gacos.py`, we intentionally dropped the support of radar-code files from `gamma/roipac` (#483), because the precise coordinates conversion from geo to radar with lookup tables in geo-coordinates (as produced by gamma/roipac) is not supported in mintpy. The initial version before #483 does support that but is very slow and complicated. I added more checkings on the input files to help clarify this.

I have been successfully run your dataset with tropo_gacos.py. Could you try it also and let me know?

yunjunz:
The `utils0.get_lat_lon_rdc()` only produce very rough lat/lon. This is fine for pyaps, as that rough lat/lon is for the interpolation of even coarser ERA5 data. Although the code itself does not give running errors, this won't work for GACOS, because it's already interpolated, and require very precise projection from GACOS spatial extent to the mintpy time-series spatial extent.

yunjunz (approved):

sssangha:
> For `tropo_gacos.py`, we intentionally dropped the support of radar-code files from `gamma/roipac` (#483), because the precise coordinates conversion from geo to radar with lookup tables in geo-coordinates (as produced by gamma/roipac) is not supported in mintpy. The initial version before #483 does support that but is very slow and complicated. I added more checkings on the input files to help clarify this.

@yunjunz thanks for the update/clarification! I now understand with roipac products it is best to proceed with the geocoded products, but is there a straightforward way to specify the use of geocoded products from the `invert_network` step and onwards? I.e., is there a way to specify the usage of geocoded products in the template file, such that you avoid having to manually pass the geocoded products in the commandline for each of the successive steps (e.g. deramping, dem error correction, velocity fit)?

Also, would these restrictions also apply to sentinel products in radar geometry (from the stack processor), or is the code designed to handle such outputs?

yunjunz:
Hi @sssangha, there is no template option for that within `smallbaselineApp.py`. I would recommend you to prepare an interferogram stack for your AOI in geo-coordinates directly, using `geocode.py --bbox --lalo`. Here (https://github.com/geodesymiami/Yunjun_etal-2021-Kirishima/blob/main/configs/load_aoi_Alos2AT131.sh) is an example where I loaded the whole track of radar-coded interferogram into HDF5/MintPy format for backup purpose, then called geocode.py to create a much smaller geocoded interferogram stack for my AOI.

For isce products, radar-coded flies are supported in `tropo_gacos.py` because of the radar-coded lookup table format, which makes the geo2radar conversion easy.

sssangha:
@yunjunz ok understood, and yes in that case the code works as expected. Outputs make sense.

PR #526, opened 2021-03-04
Track ramp coefficients

sssangha:

Update `remove_ramp.py` to produce text-files reporting function fit coefficient for given field(s). Text-file names reflect the input filename and ramp type.

For time-series files, there are n lines for each epoch with each line prepended by the epoch date.

For IFG stacks, there are n lines for each IFG with each line prepended by the IFG index.

And finally for function-fits to the TS (e.g. velocity), a single line is written out at prepended by the function-fit type.

This summary may be a bit terse, so please let me know if you’d like me to elaborate on anything.

**Reminders**

### DIFF
--- a/mintpy/objects/ramp.py
+++ b/mintpy/objects/ramp.py
@@ -19,16 +19,19 @@ RAMP_LIST = [
 ]
 
 
-def deramp(data, mask_in, ramp_type='linear', metadata=None, max_num_sample=1e6):
+def deramp(data, mask_in, ramp_type='linear', metadata=None, max_num_sample=1e6, coeff_file=None):
     '''Remove ramp from input data matrix based on pixel marked by mask
     Ignore data with nan or zero value.
-    Parameters: data      : 2D / 3D np.ndarray, data to be derampped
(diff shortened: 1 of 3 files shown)


yunjunz (changes requested):
Thank you @sssangha for the useful feature. May I suggest a few changes?

+ in `objects.ramp.deramp()`, add a 3rd output `X`, and have all text IO in `utils.run_deramp()`.

+ set the text file basename as `"rampCoeff_{}.txt".format(os.path.splitext(os.path.basename(fname))[0])`, to be consistent with the naming style in mintpy where the first part of the filename is the nature of the file.

+ add some header info to the text file, e.g. the ramp type, text content format (YYYYMMDD, yy**2, xx, 1 etc.) and use `delimiter='\t'` instead of ":" during savetxt call, here is a reference: https://github.com/insarlab/MintPy/blob/main/mintpy/objects/stack.py#L408

+ use `with open() as f:` style for file writing (but no big deal, it's up to you)

yunjunz (approved):
Look good to me now.

PR #537, opened 2021-03-19
ifgramStack.temporal_average: cast ds_size to int64 to avoid overflow 32-bit OS

### LINKED ISSUE #536: Overflow Error In Stack.py due to np.int32 value
I have noticed that there is a possible bug where it can lead to an overflow error due to `np.sum()` returning a `np.int32` object. This is fixed if it is cast as a float object.

It is in this specific line:
https://github.com/insarlab/MintPy/blob/main/mintpy/objects/stack.py#L1035

I have replicated the variables in my analysis. I am analyzing 75 interferograms using MintPy.

```python
# Parameters specific to my analysis
self.length = 1845
self.width = 1370
ds_size = np.sum(75) * self.length * self.width * 4  # ds_size is 758295000

num_step = int(np.ceil(ds_size * 3 / (4 * 1024**3))) # num_step is zero

RuntimeWarning: overflow encountered in long_scalars
```
Expected value for num_step is `1` but due to overflow error actual value is `0`. After this, the MintPy analysis fails due to `ZeroDivisionError`.

**My Solution:**
I fixed the error by temporarily casting `ds_size` as a float32 when calculating num_step.
```python
num_step = int(np.ceil(np.float32(ds_size) * 3 / (4 * 1024**3)))
```

pbrotoisworo:

Fixes potential overflow error when calculating `ds_size` by explicitly casting it has `np.int64`. `np.sum` can return `np.int32` in some cases. See #536

**Reminders**

- [x] Fix #536 

### DIFF
--- a/mintpy/objects/stack.py
+++ b/mintpy/objects/stack.py
@@ -1032,7 +1032,7 @@ class ifgramStack:
                 ref_val = dset[ifgram_flag, self.refY, self.refX]
 
             # get step size and number
-            ds_size = np.sum(ifgram_flag) * self.length * self.width * 4
+            ds_size = np.sum(ifgram_flag, dtype=np.int64) * self.length * self.width * 4
             num_step = int(np.ceil(ds_size * 3 / (max_memory * 1024**3)))
             row_step = int(np.rint(self.length / num_step / 10) * 10)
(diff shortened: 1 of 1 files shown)


yunjunz (approved):
Thank you @pbrotoisworo.

PR #538, opened 2021-03-19
platform neutral method for sbApp.plot_result()

### LINKED ISSUE #531: Add Windows functionality to plot_result function
I'm using MintPy on a Windows machine and it cannot run the `.sh` file used to generate the plots and figures. I think it would be a good idea to add this so Windows users can easily see the output. I see that in the `plot_result` function it just copies a template `.sh` file into the work directory.

I have something like this in my own personal analysis using MintPy. This could use the following format below:
```python
# Paths
pic_dir = os.path.join(self.workDir, 'pic')
dem = os.path.join(self.workDir, 'inputs', 'geometryGeo.h5')
velocity_h5 = os.path.join(self.workDir, 'velocity.h5')
mask = os.path.join(self.workDir, 'maskTempCoh.h5')

# Run view.py
view_py = f'{self.python_interpreter_path} {self.path_mintpy_view} --nodisplay --dpi 150 --update'

# Create command
cmd = f'{view_py} {velocity_h5} --dem {dem} --mask {mask} -u cm -o {os.path.join(pic_dir, "velocity.png")}'

# Run
proc = subprocess.Popen(cmd, shell=True)
```
That's just one sample but would do that for all the desired output figures. I think this can work for Linux, MacOS, and Windows systems.

I can do a pull request if interested.

pbrotoisworo:

`plot_smallbaselineApp.sh` was translated to a pure Python script. It will save all figures in the output `pic` directory, as described in #531.

Some things to note:

- Method will check if data exists before generating any plot. In my experience not all the figures get generated which leads to an error if it still tries to generate a plot with missing data.
- Only tested with Sentinel-1 data processed using SNAP. I did not test with Envisat or UNAVCO data.
- Tested on a Windows 8 machine

**Reminders**

### DIFF
--- a/mintpy/smallbaselineApp.py
+++ b/mintpy/smallbaselineApp.py
@@ -9,14 +9,12 @@
 
 
 import os
-import re
 import sys
 import glob
 import time
 import datetime
 import shutil
 import argparse
-import subprocess
 import numpy as np
 
 import mintpy
--- a/mintpy/smallbaselineApp.py
+++ b/mintpy/smallbaselineApp.py
@@ -1097,7 +1097,7 @@ class TimeSeriesAnalysis:
                 ['timeseries_*.h5'] + opt2,
                 [os.path.join(geo_dir, 'geo_maskTempCoh.h5')] + opt1 + ['-c', 'gray'],
                 [os.path.join(geo_dir, 'geo_temporalCoherence.h5')] + opt1 + ['-c', 'gray'],
-                [os.path.join(geo_dir, 'geo_velocity.h5')] + opt1 + ['velocity'],
+                [os.path.join(geo_dir, 'geo_velocity.h5'), 'velocity'] + opt1,
                 [os.path.join(geo_dir, 'geo_timeseries*.h5')] + opt1 + ['--noaxis'],
                 [f'velocity{tropo_model}.h5'] + opt1 + ['--mask', 'no'],
                 ['numInvIfgram.h5'] + opt1 + ['--mask', 'no']
@@ -1105,6 +1105,7 @@ class TimeSeriesAnalysis:
             if 'connectComponent' in readfile.get_dataset_list(ifgram_stack):
                 iargs_list.append([ifgram_stack, 'connectComponent'] + opt1 + ['--mask', 'no'])
 
+            # Convert glob paths and their arguments to lists
             for idx, iargs in enumerate(iargs_list):
                 if '*' in iargs[0]:
                     # Get the arguments of the filename
@@ -1117,6 +1118,7 @@ class TimeSeriesAnalysis:
                         for path in glob_list:
                             iargs_list.insert(idx, [path] + args)
 
+            # Generate figures
             for iargs in iargs_list:
                 if os.path.isfile(iargs[0]):
                     mintpy.view.main(iargs)
--- a/mintpy/smallbaselineApp.py
+++ b/mintpy/smallbaselineApp.py
@@ -22,6 +22,7 @@ import mintpy
 from mintpy.objects import sensor, RAMP_LIST
 from mintpy.utils import readfile, writefile, utils as ut
 from mintpy.defaults.template import STEP_LIST
+from mintpy import view
 import mintpy.workflow   # dynamic import of modules for smallbaselineApp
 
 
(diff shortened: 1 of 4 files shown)


yunjunz:
Thank you @pbrotoisworo. I would suggest using a list of lists to assemble all files and their plot configurations, such as the `iargs_list` shown below: each element list corresponds to one line in the plot_smallbaselineApp.sh file, within which the first element is the file name. 

```python
dem_file = 
mask_file = 
opt1 = ['--nodisplay', '--dpi', '150', '--update']
opt2 = ['--nodisplay', '--dpi', '150', '--update', '--dem', dem_file, '--mask', mask_file, '-u', 'cm']
iargs_list = [
    ['velocity.h5', opt1],
    ['temporalCoherence.h5', '-c', 'gray', '--vlim', '0', '1'],
    ...
]
```

Then, looping over this list to update:
1. use `glob.glob()` to search all existing file patterns: we should add `timeseries_*.h5` in the list above for shorter code, and expand this here.
2. use `os.path.isfile()` to check if file exists and remove them if not.

With the above style, we should be able to convert the `# Set paths` part into one big list `iargs_list` and to convert the `# Run coommands` part into a short for loop to call `view.main()`.

Let me know what do you think please.

pbrotoisworo:
@yunjunz I think that can work, but what do you think about this method? This time I'm trying to create a method that is intuitive and easier to read.

Consider this function to load arguments for `view.py`. It can be included within `plot_result`:
```python
            def view_input(path, view_params, iargs=None):
                "Handle plotting arguments for view.py"

                # Check if file exists. Similar to test -f $file
                if os.path.isfile(path):
                    # Path will be the first argument for view.py
                    view_params = [path] + view_params
                    if iargs:
                        # If additional iargs exist, add them to the list at the end
                        view_params += iargs
                    # Run view.py using the list
                    view.main(view_params)
```

Using that function, generating plots can be done using one line which makes the code much easier to read.

```python
            ## common view.py option for all files
            view_opt = ['--nodisplay', '--dpi', '150', '--update']

            ## Plot key files
            opt = ['--dem', dem_file, '--mask', mask_file, '-u', 'cm']

            if plot_key_files:
                file = os.path.join(self.workDir, 'velocity.h5')
                view_input(path=file, view_params=view_opt, iargs=opt)
                file = os.path.join(self.workDir, 'temporalCoherence.h5')
                view_input(path=file, view_params=view_opt, iargs=['-c', 'gray', '--vlim', '0', '1'])
```

I've done some tests using this method and it works. If this method is okay with you I can create it. Also, I did not see your initial suggestion of creating a `plot_smallbaselineApp.py`. I can try creating that file so that this plot function can be used independently and can be imported into `smallbaselineApp.py`

In `plot_smallbaselineApp.py` inputs can be included for these parameters which are available in `plot_smallbaselineApp.sh`
```
plot_key_files = 1
plot_loaded_data = 1
plot_loaded_data_aux = 1
plot_timeseries = 1
plot_geocoded_data = 1
plot_the_rest = 1
```

Please let me know what you think about this method. If you prefer a list of lists to assemble all files and their plot configurations I can do that instead.

yunjunz:
Hi @pbrotoisworo, I would still prefer the list of lists, as the workflow seems simpler.

Please discard my initial suggestions of `plot_smallbaselineApp.py`, your way of writing them inside `plot_result()` is nicer. We could change `smallbaselineApp.py --noplot` to `smallbaselineApp.py --plot` for users to run `plot_result()` only, as an alternatively of running `plot_smallbaselineApp.sh/.py`.

Both `smallbaselineApp.py --noplot` option and `plot_key_files = 1` inside plot_smallbaselineApp.sh were introduced before `view.py --update` to save time during re-run. Now they are rarely used from my experience, so we could change/remove them.

yunjunz:
Since view.py is now one of the sub-modules used in smallbaselineApp.py, I would suggest adding `view` to [workflow/\_\_inti\_\_.py#L34](https://github.com/insarlab/MintPy/blob/main/mintpy/workflow/__init__.py#L34) and use `mintpy.view.main()` to be consistent with other sub-modules.

pbrotoisworo:
@yunjunz I've updated the PR using your suggestions.

yunjunz:
`coherence`, `connectComponent` and `velcoity` are all datasets and should be immediately after the input file such as `view.py velocity.h5 velocity`.

mintpy/smallbaselineApp.py
```diff
@@ -1063,8 +1065,72 @@ def plot_result(self, print_aux=True, plot=True):
...
+                [f'timeseries_demErr.h5'] + opt2,
+                [os.path.join(geo_dir, 'geo_maskTempCoh.h5')] + opt1 + ['-c', 'gray'],
+                [os.path.join(geo_dir, 'geo_temporalCoherence.h5')] + opt1 + ['-c', 'gray'],
+                [os.path.join(geo_dir, 'geo_velocity.h5')] + opt1 + ['velocity'],
+                [os.path.join(geo_dir, 'geo_timeseries_ECMWF_demErr_ramp.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_ECMWF_demErr.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_demErr_ramp.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_demErr.h5')] + opt1 + ['--noaxis'],
+                [f'velocity{tropo_model}.h5'] + opt1 + ['--mask', 'no'],
+                [f'numInvIfgram.h5'] + opt1 + ['--mask', 'no']
```
yunjunz, line 1118: May I ask why do you add `f` to some files?

mintpy/smallbaselineApp.py
```diff
@@ -1063,8 +1065,72 @@ def plot_result(self, print_aux=True, plot=True):
...
+                ['timeseries_LODcor_ECMWF_demErr.h5'] + opt2,
+                ['timeseries_LODcor_ECMWF_ramp.h5'] + opt2,
+                ['timeseries_LODcor_ECMWF_ramp_demErr.h5'] + opt2,
+                [f'timeseries_{tropo_model}.h5'] + opt2,
+                [f'timeseries_{tropo_model}_demErr.h5'] + opt2,
+                [f'timeseries_{tropo_model}_ramp.h5'] + opt2,
+                [f'timeseries_{tropo_model}_ramp_demErr.h5'] + opt2,
+                [f'timeseries_ramp.h5'] + opt2,
+                [f'timeseries_demErr_ramp.h5'] + opt2,
+                [f'timeseries_demErr.h5'] + opt2,
```
yunjunz, line 1109: We should use the following two lines for all time-series files here.
```python
['timeseries.h5', opt2],
['timeseries_*.h5', opt2],
```

mintpy/smallbaselineApp.py
```diff
@@ -1063,8 +1065,72 @@ def plot_result(self, print_aux=True, plot=True):
...
+                [os.path.join(geo_dir, 'geo_temporalCoherence.h5')] + opt1 + ['-c', 'gray'],
+                [os.path.join(geo_dir, 'geo_velocity.h5')] + opt1 + ['velocity'],
+                [os.path.join(geo_dir, 'geo_timeseries_ECMWF_demErr_ramp.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_ECMWF_demErr.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_demErr_ramp.h5')] + opt1 + ['--noaxis'],
+                [os.path.join(geo_dir, 'geo_timeseries_demErr.h5')] + opt1 + ['--noaxis'],
+                [f'velocity{tropo_model}.h5'] + opt1 + ['--mask', 'no'],
+                [f'numInvIfgram.h5'] + opt1 + ['--mask', 'no']
+            ]
+
```
yunjunz, line 1120: And at here, loop over the iarg_list, within each loop, if * is in the file path, run glob.glob() to search all existing files, and replace the original element list with multiple element lists: one for each existing files.

mintpy/smallbaselineApp.py
```diff
@@ -1063,8 +1065,72 @@ def plot_result(self, print_aux=True, plot=True):
...
+            for iargs in iargs_list:
+                if os.path.isfile(iargs[0]):
+                    view.main(iargs)
```
yunjunz, line 1123: Since view is now in workflow, here we could use `mintpy.view.main(iargs)` and remove the `from mintpy import view` at the top.

mintpy/smallbaselineApp.py
pbrotoisworo: I used f-strings for string formatting. In your README it says MintPy works with Python 3.6+. F-strings were introduced in Python 3.6. Do you prefer I use a specific style for string formatting?

I used it like this below:
```python
# Loaded model is 'ERA5'
tropo_model = self.template['mintpy.troposphericDelay.weatherModel']

# String becomes velocityERA5.h5
f'velocity{tropo_model}.h5'
```

yunjunz: I see, this's great.

mintpy/smallbaselineApp.py
pbrotoisworo: Okay. I will fix this.

mintpy/smallbaselineApp.py
pbrotoisworo: > `unwrapPhase`, `coherence`, `connectComponent` and `velcoity` are all datasets and should be immediately after the input file such as `view.py velocity.h5 velocity` without - or --.

Thanks. This has fixed the errors except for `connectComponent`. It shows this error. I'm not sure if it's just related to my sample dataset missing this dataset.

```python
Exception: No input dataset found!
available datasets:
['unwrapPhase-20200109_20200121', 'unwrapPhase-20200109_20200226', 'unwrapPhase-20200109_20200309', 'unwrapPhase-20200121_20200226', 'unwrapPhase-20200121_20200309', 'unwrapPhase-20200121_20200321',
'unwrapPhase-20200226_20200309', 'unwrapPhase-20200226_20200321', 'unwrapPhase-20200226_20200402', 'unwrapPhase-20200309_20200321', 'unwrapPhase-20200309_20200402', 'unwrapPhase-20200309_20200414',
'coherence-20200109_20200121', 'coherence-20200109_20200226', 'coherence-20200109_20200309',
'coherence-20200121_20200226', 'coherence-20200121_20200309', 'coherence-20200121_20200321', 'coherence-20200226_20200309', 'coherence-20200226_20200321', 'coherence-20200226_20200402', 'coherence-20200309_20200321', 'coherence-20200309_20200402', 'coherence-20200309_20200414',
'wrapPhase-20200109_20200121', 'wrapPhase-20200109_20200226', 'wrapPhase-20200109_20200309', 'wrapPhase-20200121_20200226', 'wrapPhase-20200121_20200309', 'wrapPhase-20200121_20200321', 'wrapPhase-20200226_20200309', 'wrapPhase-20200226_20200321', 'wrapPhase-20200226_20200402', 'wrapPhase-20200309_20200321', 'wrapPhase-20200309_20200402', 'wrapPhase-20200309_20200414']
```

yunjunz: Currently not all supported InSAR processors provide the connectComponent data, so it's missing in your test. This only happens to the `ifgramStack.h5` file. In the .sh file, I used `h5ls` to check if the dataset exists; in python, you could check as below:

```python
if 'connectComponent' in readfile.get_dataset_list(fname):
```

pbrotoisworo:
@yunjunz kindly see the latest update to the PR. It has passed the CircleCI tests and hopefully the formatting is okay with you.

yunjunz (approved):
Looks all good to me.

PR #542, opened 2021-03-24
ingest hyp3

forrestfwilliams:

The ASF HyP3 on-demand InSAR service is a great option for creating interferograms, but MintPy is not currently set up to ingest these products. This pull request creates a prep_hyp3.py workflow that would allow for the ingestion of these products.

**Changes**
- Creates prep_hyp3.py for the creation of rsc metadata files for HyP3 products
- Adds hyp3 to load_data.py processor list, and sets up reference to prep_hyp3 when hyp3 is the processor.
- Has readfile.read process HyP3 products as GDAL tifs

Resolves Issue: #540

### DIFF
--- a/mintpy/prep_hyp3.py
+++ b/mintpy/prep_hyp3.py
@@ -2,14 +2,20 @@
 ############################################################
 # Program is part of MintPy                                #
 # Copyright (c) 2013, Zhang Yunjun, Heresh Fattahi         #
-# Author: Forrest Williams                                 #
+# Author: Forrest Williams, Mar 2021                       #
 ############################################################
+
 
 import os
 import sys
 import argparse
 from datetime import datetime
-from osgeo import gdal
+
+try:
+    from osgeo import gdal
+except ImportError:
+    raise ImportError('Can not import gdal!')
+
 from mintpy.objects import sensor
 from mintpy.utils import writefile, utils as ut
 
--- a/mintpy/prep_hyp3.py
+++ b/mintpy/prep_hyp3.py
@@ -148,7 +148,7 @@ def general_metadata(fname):
 
     # Earth radius probably won't be used anywhere for the geocoded data
     meta['EARTH_RADIUS'] = 6337286.638938101
-    
+
     return(meta)
 
 
@@ -171,14 +171,14 @@ def add_ifg_metadata(fname,meta):
     date1 = datetime.strptime(date1_string,'%Y%m%dT%H%M%S')
     date2 = datetime.strptime(date2_string,'%Y%m%dT%H%M%S')
     date_avg = date1 + (date2 - date1) / 2
-    
+
     ifg_meta_file = f'{os.path.join(directory,job_id)}.txt'
     ifg_meta = {}
     with open(ifg_meta_file, 'r') as f:
         for line in f:
             key, value = line.strip().split(': ')
             ifg_meta[key] = value
-    
+
     # Add relevant data to meta object
     meta['CENTER_LINE_UTC'] = date_avg.strftime('%y%m%d')
     meta['DATE12'] = f'{date1.strftime("%y%m%d")}-{date2.strftime("%y%m%d")}'
--- a/mintpy/prep_hyp3.py
+++ b/mintpy/prep_hyp3.py
@@ -95,103 +95,150 @@ def cmd_line_parse(iargs=None):
 
 #########################################################################
 
-# Extract geospatial and sensor metadata
-def general_metadata(fname):
-    '''Read/extract attribute data from HyP3 gdal geotiff metadata file and add to metadata dictionary
+# extract geospatial and add to meta
+def add_geospatial_metadata(fname, meta):
+    '''Read/extract attribute data from HyP3 gdal geotiff metadata and add to metadata dictionary
     Inputs:
-        *unw_phase.tif or *corr.tif file name or *dem.tif file name, e.g.
+        *unw_phase.tif, *corr.tif file name or *dem.tif file name, e.g.
             S1AA_20161223T070700_20170116T070658_VVP024_INT80_G_ueF_74C2_unw_phase.tif
             S1AA_20161223T070700_20170116T070658_VVP024_INT80_G_ueF_74C2_corr.tif
             dem.tif
+        Metadata dictionary (meta)
     Output:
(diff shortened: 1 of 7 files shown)


mintpy/prep_hyp3.py
```diff
@@ -0,0 +1,221 @@
...
+    meta["LON_REF1"] = x0
+    meta["LON_REF2"] = x1
+    meta["LON_REF3"] = x0
+    meta["LON_REF4"] = x1
+
+    meta["LAT_REF1"] = y0
+    meta["LAT_REF2"] = y0
+    meta["LAT_REF3"] = y1
+    meta["LAT_REF4"] = y1
+
```
yunjunz, line 143: `LAT/LON_REF1/2/3/4` are in a different order between ascending and descending datasets, as shown here: https://github.com/yunjunz/MintPy/blob/main/mintpy/prep_gmtsar.py#L103.

Could you please also fix this bug in `prep_aria.py` here?

mintpy/prep_hyp3.py
```diff
@@ -0,0 +1,221 @@
...
+    ifg_meta_file = f'{os.path.join(directory,job_id)}.txt'
+    ifg_meta = {}
+    with open(ifg_meta_file, 'r') as f:
+        for line in f:
+            key, value = line.strip().split(': ')
+            ifg_meta[key] = value
+
+    # Add relevant data to meta object
+    meta['CENTER_LINE_UTC'] = date_avg.strftime('%y%m%d')
```
yunjunz, line 187: This attribute contains the time of the day only, in the unit of seconds, without date info. Here (https://github.com/insarlab/MintPy/blob/main/mintpy/utils/isce_utils.py#L159) is an example.

yunjunz:
This is awesome @forrestfwilliams! I only have a few minor suggestions to complete this support:

1. add HyP3 with link to the `docs/README`
2. add hyp3 to the comment of `mintpy.load.processor` in `mintpy/defaults/smallbaselineApp.cfg`
3. could you add an example directory structure and its template options to `docs/dir_structure.md`? That will be a good place to add some descriptions and/or links on the tutorials as well if available.
4. if it's not big trouble, could you please send me a test dataset for me to try it as well?

mintpy/prep_hyp3.py
forrestfwilliams: Just to be clear, would you like me to fix this bug for both prep_hyp3 and prep_aria in this PR?

forrestfwilliams:
Hi @yunjunz I'll get started on these changes! What is the best way to get the test dataset to you?

mintpy/prep_hyp3.py
forrestfwilliams: Also, should this change also be made for the geometry (DEM) dataset as well with the ascending/descending information from one of the interferograms? Out of curiosity, why is the order different for ascending and descending datasets?

yunjunz: > Just to be clear, would you like me to fix this bug for both prep_hyp3 and prep_aria in this PR?

Yes please.

yunjunz: > should this change also be made for the geometry (DEM) dataset as well with the ascending/descending information from one of the interferograms? 

Yes, it's should be in the metadata of all files.

From the strict [definition of REF_LAT/LON1/2/3/4](https://mintpy.readthedocs.io/en/latest/api/attributes/), they are in the order of first line near/far range, then last line near/far range, thus, for the same area, they are different between asc and desc.

mintpy/prep_hyp3.py
yunjunz: float

yunjunz:
Hi @forrestfwilliams, I just sent you a link to my Box; you could also upload it to Google Drive if you have enough space there.

forrestfwilliams:
Hi @yunjunz, I've uploaded test data to the box drive, it should match the directory structure I've described in `docs/dir_structure.md`. I've only included two interferograms to decrease the file size, but let me know if you would like more.

yunjunz:
Thank you @forrestfwilliams, the doc looks great.

For the test data, could you send me a few more to have some redundacy? You have sent me date1_date3, date2_date3, maybe add date1_date2?

forrestfwilliams:
@yunjunz I've made all the changes we've discussed and the updated example data is in the box folder!

yunjunz:
@forrestfwilliams In the newly added 1st interferogram, I did not find the `*_clip.tif`, could you make sure?

We need `incidence angle` and `slant range distance` for tropospheric delay correction and DEM error correction. The ideal case would be the pixel-wised 2D matrix. If they are not available, a mean value also works, in this case, could you add the following two attributes for it? They will be used by `geometryDict` during load_data.py to write the mean 2D matrix.

```
SLANT_RANGE_DISTANCE =   # mean slant range distance between the satellite and the ground
INCIDENCE_ANGLE =   # mean incidence angle in degree
```

yunjunz (approved):
This PR itself works fine without errors, so I think it's ready to be merged. Thank you @forrestfwilliams!!!

There are a few more things to consider, to better support the full functionalities of MintPy. Maybe for future PRs, unless you want to do it in this PR? @forrestfwilliams Let me know please.

1. In the newly added 1st interferogram, I did not find the `*_clip.tif`, could you make sure @forrestfwilliams?

2. We need `incidence angle` and `slant range distance` for tropospheric delay correction and DEM error correction. The ideal case would be the pixel-wised 2D matrix. If they are not available, a mean value also works, in this case, could you add the following two attributes for it? They will be used by `geometryDict` during load_data.py to write the mean 2D matrix. 

```
SLANT_RANGE_DISTANCE =   # mean slant range distance between the satellite and the ground
INCIDENCE_ANGLE =   # mean incidence angle in degree
```

3. The "meter" unit is not properly displayed in view.py currently, which requires some changes (I will probably try it sometime this weekend or next week). For now, one could run `view.py --coord radar` to force the display in radar coordinate.

forrestfwilliams:
@yunjunz 1. was my mistake, sorry. I've uploaded the correct version of the example data to the box folder. As for 2. and 3. let's fix them in a future PR. `incidence angle` and `slant range distance` are not available in the current version of HyP3, but the project is still in Beta, so they'll likely be open to adding this data as an optional output. Let's go ahead and merge!

PR #543, opened 2021-03-28
save_gbis: maskout pixels w zero phase value

bjmarfito:
My officemate noticed that the output using save_gbis.py contains 0 values in waterbody areas and areas masked out using the connected component file from the geo_velocity.h5 file. GBIS requires 0 and NaN values be removed on the .mat file before doing the inversion.

The top figure shows the 0 values (waterbody) indicated as red areas while the bottom figure shows the removal of the 0 values after applying the modification on the save_gbis.py.

with_0

without_0

*Added lines of code to remove 0 and NaN values prior to exporting to the .mat file.

**Reminders**

### DIFF
--- a/mintpy/save_gbis.py
+++ b/mintpy/save_gbis.py
@@ -218,12 +218,16 @@ def save2mat(inps):
     mdict['Mask'] = inps.mask
     mdict['Metadata'] = inps.metadata
     mdict['Phase'][mdict['Phase']==0] = np.nan
+    placeholderRef = np.where((mdict['Lat'] == inps.metadata['REF_LAT']) & (mdict['Lon']==inps.metadata['REF_LON']))
+    mdict['Phase'][placeholderRef] = 0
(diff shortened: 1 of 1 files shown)


yunjunz:
`save_gbis.py` supports masking via the `-m / --mask` option, where you could feed the connected component file for example. 

Masking out zero phase value also sounds fine with me. I would suggest:
1. keep the reference pixel, which always has zero value
2. printout a msg showing how many pixels are masked out due to zero phase value, so that users are aware of this potential observation reduction, since this operation is applied all the time.

bjmarfito:
Thank you @yunjunz for your comments. I've edited it so the reference pixels will have 0 value. Pls. check it if I've done it correctly. I've also added your suggestion to display the remaining pixels after converting it to GBIS format.

yunjunz (approved):

yunjunz:
I moved the masking operation from `save2mat()` to `read_data()`, where the other mask is done.

PR #549, opened 2021-04-22
add save_kite.py for pyrocko suite

mgovorcin:
save_kite.py
 - save velocity, step, or timeseries displ into KITE format (yaml, npz) 
 - needs kite and pyrocko module to work.: https://pyrocko.org/ 
 - prepared data can be used for inversions in GROND or BEAT
 - did not try it with unwrapped ifgs, maybe needs a few edits
 
small fix in asc_desc ..
 - converted azimuth angle from degree to radians in the design matrix

**Reminders**

### DIFF
--- a/mintpy/save_kite.py
+++ b/mintpy/save_kite.py
@@ -10,7 +10,7 @@ import sys
 import argparse
 import datetime as dt
 import numpy as np
-from mintpy.utils import readfile
+from mintpy.utils import ptime, readfile
 
 try:
     from kite.scene import Scene, SceneConfig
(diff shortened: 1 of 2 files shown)


yunjunz (changes requested):
Thank you @mgovorcin for this useful contribution!

Your `asc_desc2horz_vert.py` fix looks good to me.

As for `save_kite.py`, both Codacy and Circle CI flagged some issues. Could you fix them? I believe rebaseing your branch to the latest insarlab/main should fix the Circle CI test issue.

mintpy/save_kite.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+d2r = np.pi / 180.
+r2d = 180. / np.pi
+
+
+EXAMPLE = """example:
+  #for Velocity [use step to export coseismic displacement]:
+  save_kite.py geo/geo_velocity.h5 -g geo/geo_geometry.h5
+  save_kite.py geo/geo_velocity.h5 -g geo/geo_geometry.h5 -m geo/geo_maskTempCoh.h5
```
yunjunz, line 28: In this case, the code will read linear velocity, which has the unit of m/year. Is this expected?

For other modeling code, e.g. GBIS, we usually convert the velocity to the corresponding displacement using DATE12, so that only the displacement in the unit of m is used.

yunjunz:
@falkamelung and @ranneylxr, could you please take a look at this PR? Marin added the long-desired pyrocko support. The code itself looks good to me, but I don't have pyrocko installed for real testing.

yunjunz (approved):

mintpy/save_kite.py
mgovorcin: BEATv1.1 has a possibility to model interseismic velocities and tensile dislocations, so in those cases, this export would be needed. Example: https://pyrocko.org/beat/docs/current/examples/Rectangular_tensile.html

But for other event-type inversions, we need the displacements. Maybe would be good to add a WARNING/REMINDER message at the end of the code, if someone exports velocities accidentally.

ranneylxr:
Hi Yunjun @yunjunz, thank you for inviting me. The code looks good to me based on my experience about pyrocko.

mintpy/save_kite.py
yunjunz: Thank you @mgovorcin for the explanation, these sound great, I will add more comments on the examples. We could set `--dset` option as required, to force the user to think about this before blindly running.

yunjunz:
Thank you @ranneylxr.

PR #552, opened 2021-04-30
spatial_filter: add double difference filter option

forrestfwilliams:
The double difference filter as used in Bekaert et al 2020 has been shown to effectively filter out non-local modes of deformation. This particularly when using MintPy to monitor local forms of deformation, such as landslides. This commit implements the double difference filter in spatial_filter.py and reformats the filter_par parameter so that it accepts multiple arguments. Also fixes bug that caused ONLY the datasets specified to be written to the output file when inputs datasets are specified.

**Reminders**

### DIFF
--- a/mintpy/spatial_filter.py
+++ b/mintpy/spatial_filter.py
@@ -184,7 +184,7 @@ def filter_file(fname, ds_names=None, filter_type='lowpass_gaussian', filter_par
     ds_all = readfile.get_dataset_list(fname)
     if not ds_names:
         ds_names = ds_all
-    ds_skips = set(ds_all)-set(ds_names)
+    ds_skips = list(set(ds_all) - set(ds_names))
 
     maxDigit = max([len(i) for i in ds_names])
     dsDict = dict()
--- a/mintpy/spatial_filter.py
+++ b/mintpy/spatial_filter.py
@@ -162,10 +162,17 @@ def filter_file(fname, ds_names=None, filter_type='lowpass_gaussian', filter_par
                                      os.path.splitext(fname)[1])
 
     # filtering file
+    ds_all = readfile.get_dataset_list(fname)
     if not ds_names:
-        ds_names = readfile.get_dataset_list(fname)
+        ds_names = ds_all
+    ds_skips = set(ds_all)-set(ds_names)
+
     maxDigit = max([len(i) for i in ds_names])
     dsDict = dict()
+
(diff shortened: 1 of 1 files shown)


yunjunz:
This is cool. Could you add more detailed reference info for the Bekaert et al., 2020 paper? [Here](https://github.com/insarlab/MintPy/blob/main/mintpy/timeseries2velocity.py#L39) is an example for a reference. More detailed comments, e.g. the equation numbers, would be even better.

dbekaert:
Great. @forrestfwilliams could you perhaps give me a runfile config example for me to test?

forrestfwilliams:
Sounds good @yunjunz and @dbekaert, I'll get started on these requests!

forrestfwilliams:
@bekeart, I've worked up an example using the Sentinel-1 on Fernandina with ISCE example data:

1. Download example data
```
wget https://zenodo.org/record/3952953/files/FernandinaSenDT128.tar.xz
tar -xvJf FernandinaSenDT128.tar.xz
cd FernandinaSenDT128/mintpy
```

2. Load data
```
smallbaselineApp.py ${MINTPY_HOME}/mintpy/data/input_files/FernandinaSenDT128.txt --end load_data
```

3. Run double_difference filter and replace ifgramStack.h5 with result
```
cd inputs
spatial_filter.py  ifgramStack.h5 unwrapPhase -f double_difference -p 5 10
rm ifgramStack.h5
mv ifgramStack_double_difference.h5 ifgramStack.h5
cd ..
```

4. Run MintPy as normal from there
```
smallbaselineApp.py ${MINTPY_HOME}/mintpy/data/input_files/FernandinaSenDT128.txt --start modify_network
```

You can also perform any masking you desire before running spatial_filter.py. Hope this helps!``

yunjunz (approved):
Thank you @forrestfwilliams! It looks good to me. Testing over a velocity runs without issue. Let's see if @dbekaert has any further suggestions.

dbekaert (approved):
Approach looks correct implemented: time-series analysis on the unwrapped IFGS which we generated from regional filter-local filter. Im will pull that data to run your example, but this all looks good to me.

dbekaert:
@forrestfwilliams Thanks for making it more broadly available to others!

forrestfwilliams:
@yunjunz and @dbekaert, thank you! Looking forward to implementing this approach in my own work!

PR #559, opened 2021-05-08
stackDict.py, readfile.py: follow-up edit of passing geom from tmpl

yuankailiu:
Please review the following changes made on `stackDict.py` and `readfile.py`. This would help to compute slant range distance when loading geometry files from topsApp geometry products. 

+ stackDict.py
    - Generate 2D slant range distance if missing from input template file:
        For dataset in geo-coordinates, use:
            1) incidenceAngle matrix if available OR
            2) constant value from SLANT_RANGE_DISTANCE.

+ readfile.py
    - binary filename for los.rdr or los.geo, ... is now more flexible in read_binary_file(): filenames start with 'los' is allowable
    - force to use the specified metadata file in read_attribute()

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/mintpy/objects/stackDict.py
+++ b/mintpy/objects/stackDict.py
@@ -368,15 +368,21 @@ class geometryDict:
             return None
 
         if 'Y_FIRST' in self.extraMetadata.keys():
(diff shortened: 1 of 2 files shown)


yunjunz (approved):
Thank you @yuankailiu.

PR #563, opened 2021-05-22
create mst after selection by other thresholds

bbuzz31:

Calculate minimum spanning tree from network (based on average coherence) after dropping interferograms based on other user criteria.

Previously, the minimum spanning tree was first created from all possible interfograms. Subsequently, interferograms would be dropped from the network based on user specifications (e.g., average temporal coherence), potentially breaking the network.

Now, the minimum spanning tree is calculated after all other considerations (again, e.g. average temporal coherence) from the remaining network. Note this doesn't guarantee the network will be connected, and also does not apply to the `manual` option for dropping network nodes.

More details can be found at the discussion [here](https://groups.google.com/g/mintpy/c/1JXBYEo8iB4).

**Reminders**


yunjunz (approved):
Thank you @bbuzz31!

PR #587, opened 2021-06-22
small changes: dem_error, ifgram_inversion, plot_network

yuankailiu:

1. A Romanian-flag colormap is added into the path: `mintpy/data/colormaps/`. The colormap goes from red to yellow, to blue. One example here is with the [network plot with continuous colormap](https://ibb.co/qp2bZFd). It's showing high coherence (blue), intermediate level coherence (yellow), and low coherence (red). I guess this will be useful.

2. The current DEM error correction has polynomial and step functions. Now we can add periodic time functions since the `timeseries2velocity.py` code is much advanced. Maybe we could just move the generation of timeseriesResidual file from dem_error.py to ts2vel.py eventually. But I don't know how to do it right now.

3. ifgram_inversion.py: Now it allows for inverting the time series of the ionosphere correction. The code now can recognize the corresponding head of the filename (i.e., starting with "ion").

4.  plot_network.py: I think it would be good to make the baseline plot with zero-mean perpendicular baselines. So that the entire plot looks centered around zero, rather than systematically shifted upward or downward.

**Reminders**


mintpy/data/colormaps/romanian.cpt
```diff
@@ -0,0 +1,261 @@
...
+0.972549 209 29 37 0.976471 208 28 37
+0.976471 208 28 37 0.980392 208 26 37
+0.980392 208 26 37 0.984314 208 24 37
+0.984314 208 24 37 0.988235 207 23 37
+0.988235 207 23 37 0.992157 207 21 38
+0.992157 207 21 38 0.996078 207 19 38
+0.996078 207 19 38 1.000000 206 18 38
+N 255 255 255
+B 0 44 223
+F 206 18 38
```
yunjunz, line 261: Could you add this custom colormap name to the documentation, e.g. here (https://github.com/insarlab/MintPy/blob/main/docs/api/colormaps.md#colormaps-from-cpt-city) if it's from cpt-city?

mintpy/dem_error.py
```diff
@@ -76,6 +76,8 @@ def create_parser():
     parser.add_argument('-s', '--step-date', dest='stepFuncDate', nargs='*', default=[],
                         help='Date of step jump for temporal deformation model,'+
                              ' i.e. date of earthquake/volcanic eruption')
+    parser.add_argument('--periodic', dest='period', type=float, nargs='+', default=[],
+                        help='periodic functinos of temporal deformation model, default = None')
```
yunjunz, line 80: Adding this sounds fine to me. Could you use `dest='periodic'` to be consistent with `timeseries2velocity.py`?

yunjunz: Regarding the help msg, here the default is `[]`, not None. I would recommend using `(default: %(default)s)` style to avoid this kind of typo.

yuankailiu: No problem! Thanks! I will update these suggestions.

mintpy/ifgram_inversion.py
```diff
@@ -189,6 +189,9 @@ def cmd_line_parse(iargs=None):
         elif inps.obsDatasetName.startswith('rangeOffset'):
             inps.outfile = ['timeseriesRg.h5', 'residualInvRg.h5', 'numInvOffset.h5']
 
+        elif inps.obsDatasetName.startswith('ion'):
+            inps.outfile = ['timeseriesIon.h5', 'residualInvIon.h5', 'numInvIon.h5']
+
```
yunjunz, line 194: for ionosphere, it's still phase, so we should use `temporalCoherenceIon.h5` instead of `residualInvIon.h5`.

yuankailiu: I see. I will correct that. Thanks.

yunjunz (approved):
Thank you @yuankailiu. All looks good and helpful.

I agree that we should move the timeseriesResidual file to ts2vel.py. Let's create an independent issue for that.

PR #599, opened 2021-07-08
stackDict: boolean conversion of waterBody to waterMask

yuankailiu:
**Overview**

I think the `data` read from a waterBody file is a boolean array with water=True; land=False

We only need to flip it to make a waterMask where water=False; land=True

I have a waterBody file generated from isce, which looks like below
land as 0; water as -1
image

When loading into MintPy, it converts to a waterMask (land as 1; water as 0)

But since the `dsDataType` has been [set to np.bool_ at here](https://github.com/insarlab/MintPy/blob/main/mintpy/objects/stackDict.py#L579-L580) when the dataset is waterMask
So [as the waterBody is read here](https://github.com/insarlab/MintPy/blob/main/mintpy/objects/stackDict.py#L589-L596), the `data` is already a boolean 2d array (True/False) where water=True (original value=-1) and land=False (original value=0)

That’s what my terminal output shows based on testing on printing out the `data` array and the sum of True values of `data`.
image

image

So as the code does `data = data > -0.5`, everything is converted to be True. Then I end up getting all ones in my waterMask

I propose by changing [this line](https://github.com/insarlab/MintPy/blob/main/mintpy/objects/stackDict.py#L599) to the following can make a proper waterMask file
```python
                        data = ~data
```

Thanks!

**Reminders**

### DIFF
--- a/mintpy/objects/stackDict.py
+++ b/mintpy/objects/stackDict.py
@@ -596,7 +596,7 @@ class geometryDict:
                     # water mask:  0 for water and 1 for land
                     fname = os.path.basename(self.datasetDict[dsName])
                     if fname.startswith('waterBody') or fname.endswith('.wbd'):
-                        data = data > -0.5
+                        data = ~data
                         print(('    input file "{}" is water body (-1/0 for water/land), '
                                'convert to water mask (0/1 for water/land).'.format(fname)))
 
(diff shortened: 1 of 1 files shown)


yunjunz (approved):
This makes sense, thank you @yuankailiu.

PR #608, opened 2021-07-14
fix bug in aligning ifgs with coherence

bbuzz31:
There was a small bug (likely introduced by me) with an important impact in coherence based network modification. There was a set difference that caused a mismatch between an interferogram and it's coherence, which then gets propagated through the program. Simply sorting the set difference solves the problem.


yunjunz (approved):
Thank you for the fix @bbuzz31.

yunjunz:
This bug should affect the network modification result by not getting the exact network as intended from the configuration of `mintpy.network.coherenceBased = yes` (no by default) AND `mintpy.network.keepMinSpanTree = yes` (yes by default). However, this does not affect the correctness of the processed result.

PR #610, opened 2021-07-15
generate_mask: mask pixel clusters by size

forrestfwilliams:
When looking for local deformation signals in MintPy results, it is useful to mask the output velocity.h5 by a quality metric (i.e. temporal coherence) then identify clusters of pixels that are larger than a certain size threshold. This pull request would add functionality to `generate_masks.py` which would enable the user to mask out pixel clusters that contain less than a given number of pixels.

**Changes**
- generate_mask
    - adds minpixels (-p) argument
    - adds pixel cluster size masking

**Reminders**

### DIFF
--- a/mintpy/generate_mask.py
+++ b/mintpy/generate_mask.py
@@ -25,6 +25,9 @@ EXAMPLE = """example:
   # exlcude area by min/max value and/or subset in row/col direction
   generate_mask.py  081018_090118.unw -m 3 -M 8 -y 100 700 -x 200 800 -o mask_1.h5
 
+  # exlcude pixel cluster based on minimum number of pixels
+  generate_mask.py  maskTempCoh.h5 -p 10 mask_1.h5
+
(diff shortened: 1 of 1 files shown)


yunjunz:
Thank you @forrestfwilliams, I think this is a useful new feature.

Have you tried [`skimage.morphology.remove_small_objects()`](https://scikit-image.org/docs/stable/api/skimage.morphology.html#skimage.morphology.remove_small_objects) for this purpose? 

I used this skimage function while handling the connected components in [mintpy.objects.conncomp.py#L40](https://github.com/insarlab/MintPy/blob/main/mintpy/objects/conncomp.py#L40) and it has been a simple and pleasant experience.

forrestfwilliams:
Thank you @yunjunz, I wasn't aware of this function. I'll rewrite the PR using it!

yunjunz (approved):
Looks great, thank you @forrestfwilliams.

PR #615, opened 2021-07-25
read_attr: check existing FILE_TYPE for HDF5 file

mirzaees:

When I run info.py, the "FILE_TYPE" is guessed based on dataset name and if a stack with different naming is called, some options like --date do not work. 
This change proposes to read 'FILT_TYPE' from the stack attributes if exists.

**Reminders**

### DIFF
--- a/mintpy/utils/readfile.py
+++ b/mintpy/utils/readfile.py
@@ -709,9 +709,7 @@ def read_attribute(fname, datasetName=None, metafile_ext=None):
 
         # FILE_TYPE - k
         py2_mintpy_stack_files = ['interferograms', 'coherence', 'wrapped'] #obsolete mintpy format
-        if 'FILE_TYPE' in atr:
-            k = atr['FILE_TYPE']
-        elif any(i in d1_list for i in ['unwrapPhase', 'azimuthOffset']):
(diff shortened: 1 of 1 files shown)


mintpy/utils/readfile.py
```diff
@@ -709,7 +709,9 @@ def read_attribute(fname, datasetName=None, metafile_ext=None):
 
         # FILE_TYPE - k
         py2_mintpy_stack_files = ['interferograms', 'coherence', 'wrapped'] #obsolete mintpy format
-        if any(i in d1_list for i in ['unwrapPhase', 'azimuthOffset']):
+        if 'FILE_TYPE' in atr:
+            k = atr['FILE_TYPE']
```
yunjunz, line 713: Could we move this here before using the random group/dataset names? That won't change the behavior too much, as I want the function to be less sensitive to the potential wrong FILE_TYPE attribute, which I found myself easily cause when creating an h5 file without specifying the FILE_TYPE.

https://github.com/insarlab/MintPy/blob/9e828af2cecac0f5f05b347a596b4b2736ca1046/mintpy/utils/readfile.py#L728

mirzaees: Yes, for sure. I'll change it

yunjunz (approved):
LGTM, thank you @mirzaees.

PR #624, opened 2021-08-02
ts2vel: skip writing for empty blocks

olliestephenson:

Fixing a minor bug  - write_hdf5_block function comes from the writefile import


olliestephenson:
I've realised that the write_hdf5_block function from writefile needs different arguments, so my bug fix isn't valid

yunjunz:
yeah, it is from an older version of timeseries2velocity.py which has its own layout_hdf5/write_hdf5_block(). The bug is still valid. Thank you for pointing this out.

It turns out that we don't need to write for empty blocks at all, since the file has been initiated with default values.

yunjunz (approved):

PR #634, opened 2021-08-10
asc_desc2horz_vert --oo: grab sensor name from PROJECT_NAME

olliestephenson:
Passing the project name, rather than the file path, to project_name2sensor_name seems like a more reliable way of getting the sensor name.


yunjunz (approved):
By default, PROJECT_NAME is grabbed from FILE_PATH or template file name in load_data.py. However, one could manually add/modify PROJECT_NAME afterward. So I agree with you.

PR #635, opened 2021-08-10
asc_desc2horz_vert.py --oo: Avoid overwriting modified attributes

olliestephenson:
Currently the file attributes are modified in the `asc_desc_files2horz_vert` function to take account of the fact that we're looking at the overlap between two tracks. However the `write_to_one_file` as currently written overwrites the input `atr` variable when looping over the list of attributes for the input files. The least to the attributes of the newly created file being incorrect. I've proposed a renaming to avoid this.


yunjunz (approved):
Good catch! Thank you @olliestephenson.

PR #636, opened 2021-08-10
asc_desc2horz_vert: bugfix on custom azimuth and refactor

olliestephenson:
The azimuth variable is supplied as part of the command line input, however it's not actually passed to the `get_design_matrix` function where its needed, meaning the command line variable does nothing and the azimuth angle defaults to 90 degrees. I've added code to pass it through to the design matrix function, so now the azimuth angle can be modified via the command line.

### DIFF
--- a/mintpy/asc_desc2horz_vert.py
+++ b/mintpy/asc_desc2horz_vert.py
@@ -79,7 +79,7 @@ def cmd_line_parse(iargs=None):
     # check input azimuth angle
     if inps.azimuth < 0.:
         inps.azimuth += 360.
-    inps.azimuth *= np.pi/180.
+    #inps.azimuth *= np.pi/180.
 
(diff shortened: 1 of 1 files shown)


yunjunz:
I find my notations in the script are not very clear, so I refactor them here. Below is the notebook I used for testing:
https://github.com/insarlab/MintPy-tutorial/blob/main/applications/asc_desc2horz_vert.ipynb

yunjunz:
For anyone who might be interested or concerned, the version before this PR works fine with the default azimuth direction. So I don't expect any impact if you did not specify an azimuth angle in the command line.

yunjunz:
Hi @olliestephenson, please note that I change the input azimuth angle definition to `anti-clockwise` as positive, still measured from the north, to be consistent with the rest of the mintpy and isce-2 convention.

PR #639, opened 2021-08-13
tropo_pyaps3: check GAM account info in pyaps3/model.cfg file

### LINKED ISSUE #226: Catch and report PyAPS Copernicus Climate Store credential error
Not sure if it is bug report or feature request.

ERA5 PYAPS Tropospheric delay correction currently fails with a hard to diagnose ZeroDivisionError in the progress bar class when the user's login credentials for _https://cds.climate.copernicus.eu/_ is not correct (or if the user has not accepted the T&Cs). This makes a simple problem (user registration) very hard to figure out.

Going through the code and error stack below reveals that the division error is a result of the _date_list2grib_file_ function not returning a list of data since it does not have access to the service. The progress bar then uses the length of this empty list to set the endpoint of the progress bar resulting in the error. 

I think one solution would be to catch the specific error in the _try_ clause (line 398) of the dload_grib_files function in tropo_pyaps3.py. 

> calcualting` delay for each date using PyAPS (Jolivet et al., 2011; 2014) ...
number of grib files used: 0
  File "/home/app/app.py", line 10, in <module>
    smallbaselineApp.main()
  File "/home/python/MintPy/mintpy/smallbaselineApp.py", line 1069, in main
    app.run(steps=inps.runSteps, plot=inps.plot)
  File "/home/python/MintPy/mintpy/smallbaselineApp.py", line 1019, in run
    self.run_tropospheric_delay_correction(sname)
  File "/home/python/MintPy/mintpy/smallbaselineApp.py", line 751, in run_tropospheric_delay_correction
    tropo_pyaps3.main(scp_args.split())
  File "/home/python/MintPy/mintpy/tropo_pyaps3.py", line 570, in main
    get_delay_timeseries(inps, atr)
  File "/home/python/MintPy/mintpy/tropo_pyaps3.py", line 523, in get_delay_timeseries
    prog_bar = ptime.progressBar(maxValue=num_date)
  File "/home/python/MintPy/mintpy/utils/ptime.py", line 277, in __init__
    self.reset()
  File "/home/python/MintPy/mintpy/utils/ptime.py", line 282, in reset
    self.update_amount(0)  # Build progress bar string
  File "/home/python/MintPy/mintpy/utils/ptime.py", line 296, in update_amount
    percentDone = (diffFromMin / np.float(self.span)) * 100.0
ZeroDivisionError: float division by zero

pbrotoisworo:

This addresses the error of missing API config for PyAPS. If the user did not input their API credentials in `model.cfg` in the `pyaps3` directory it raised a vague `ZeroDivisionError: float division by zero` error.

This revision will now raise a `ValueError` with a more appropriate error message telling users that the API config is missing.

Related to #226.

**Reminders**

### DIFF
--- a/mintpy/tropo_pyaps3.py
+++ b/mintpy/tropo_pyaps3.py
@@ -6,6 +6,7 @@
 ############################################################
 
 
+from configparser import ConfigParser
 import os
 import sys
 import re
--- a/mintpy/tropo_pyaps3.py
+++ b/mintpy/tropo_pyaps3.py
@@ -524,7 +524,7 @@ def check_exist_grib_file(gfile_list, print_msg=True):
 
 
 def check_pyaps_config(tropo_model):
-    """Check for input in PyAPS config file. If there is no input an error will be raised.
+    """Check for input in PyAPS config file. If they are default values or are empty then raise error
     Parameters: tropo_model : String of tropo model being used to calculate tropospheric delay
     Returns:    None
     """
@@ -537,6 +537,13 @@ def check_pyaps_config(tropo_model):
         'NARR': 'NARR'
     }
 
+    # Default values in cfg file
+    default_values = ['the-email-address-used-as-login@ecmwf-website.org',
+                      'the-user-name-used-as-login@earthdata.nasa.gov',
+                      'the-password-used-as-login@earthdata.nasa.gov',
+                      'the-email-adress-used-as-login@ucar-website.org',
+                      'your-uid:your-api-key']
+
     # Load API keys
     cfg_path = os.path.join(os.path.dirname(pa.__file__), 'model.cfg')
     cfg = ConfigParser()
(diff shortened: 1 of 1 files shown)


pbrotoisworo:
Something I just noticed now:  According to this [L116 in tropo_pyaps3.py](https://github.com/insarlab/MintPy/blob/83a2fb187699879dc208ab6ef3fa2f11ef274e1c/mintpy/tropo_pyaps3.py#L116) you only accept ERA5.

The PR checks all the tropo models that were in `model.cfg` (ERA, MERRA, ECMWF, etc). If you are not accepting other models maybe the `model.cfg` can be cleaned up for clarity. I can also revise the code to remove checks for unused tropo models.

yunjunz:
Thank you @pbrotoisworo. I just cleaned up the `model.cfg` a little bit following your suggestion (https://github.com/yunjunz/PyAPS/commit/3ba82ff537d12d24c9bf594fc985eae21f94bfac).

For the checking here, either removing the other models or keeping them is good to me. If we keep all of them, it will be good to move this checking to an independent function, and call it here.
Below is for your reference:
[CDS] -> ERA5
[ECMWF] -> ERAI
[MERRA] -> MERRA
No checking is required for NARR.

pbrotoisworo:
@yunjunz. I've created a separate function to check the pyaps config and kept the other tropo models to prevent errors if people want to use them in the future.

The function will check for default values (e.g `your-uid:your-api-key`) and empty values for the selected tropo model.

yunjunz (approved):
Looks great. Thank you @pbrotoisworo for addressing this issue!

PR #647, opened 2021-08-26
Add doc on install MintPy on Windows

Jlrine2:
These changes add instructions for installing MintPy using Conda on Windows. It also fixes a few bugs that allow windows to run through the examples in the Readme.md file of this repo.

There are still likely bugs as I haven't done any extensive testing on windows.

**Reminders**

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,6 +1,6 @@
 ## Install MintPy
 
-Tested on macOS and Linux, not sure about Windows.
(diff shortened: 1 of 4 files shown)


yunjunz:
> Yes, that will work too, except the requirements.txt is missing gdal>=3 and the pip installs so users will have to add those. Do you want us to write instructions for both like the current via conda instructions?

Great, yes please. That would be more flexible on the user side.

yunjunz (approved):

Issue #652, opened 2021-08-30
mintpy adjustments for conda-forge


yunjunz:
Copy over the to-do list in mintpy to simplify the conda-forge recipe by @jhkennedy from #648:

* [ ] `test_smallbaselineApp.py` and similar should look for configs relative to itself, not `$MINTPY_HOME` or `mintpy.__file__`
* [ ] not printing on import: https://github.com/insarlab/MintPy/blob/b307f962c5b61b58ab6864a32d3d42cf144a1c99/mintpy/__init__.py#L18-L22
* [ ] Drop any reliance on manipulating environment variables for installation (`PATH`, `PYTHONPATH` especially; `MINTPY_HOME` as well ideally) by leveraging `console_scripts` `entry_points` in `setup.py`

yunjunz:
@jhkennedy could you expand more on the 3rd topic? 

The code does not rely on the `MINTPY_HOME` environment variable, I made sure of that a while ago. So it should be trivia to remove from the installation note if needed; I kept it there for convenience purposes only.

jhkennedy:
@yunjunz yep! I'll expand soon; I'm just waiting for the conda-forge PR to go through so I can point to specific things in the feedstock.

jhkennedy:
So overall, I'm coming from the perspective of "I want my build/installation tool to do as much of the work as possible." This lets the tool handle all the different OS/machine nuances and me to focus just on telling it want I want it to do. 

MintPy, nicely, *could* be built/installed entirely by `pip`, and *could* qualify as a [noarch](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-builds) python package on conda-forge with the above changes. (meaning itdoesn't need OS special handling and therefore only needs a single build of Mintpy instead of a python-version-by-supported-OS matrix of builds.)

To illustrate, I'll go over how the [current MintPy v1.3.1 recipe](https://github.com/jhkennedy/staged-recipes/tree/add-mintpy/recipes/mintpy) doesn't qualify because it fails these criteria:

1.  No activate scripts
2.  No OS-specific build scripts -- this is tightly coupled to
    *  `scripts` argument in `setup.py` is not used
    * If `console_scripts` `entry_points` are defined in `setup.py` or `setup.cfg`, they are also listed in the `build` section of `meta.yaml`
    
    because MintPy expects things to be run on the command line/console

### 1. No activate scripts

The current activate script does two things.
1. Exports `MINTPY_HOME`
[...]

PR #653, opened 2021-09-01
__init__: rm `sys.path.insert()` and force internal imports

yunjunz:

This PR address the 1st and 2nd issues in #652.

+ `mintpy/__inti__.py`: 
   - remove try/except for print and MINTPY_HOME as it's not really used anywhere, and their corresponding module imports.
   - remove `sys.path.insert()` calls. These were replicates from the isce-2 style (isceobj, iscesys, etc.), but they are not really applicable to and used in mintpy, nor had we encouraged in any other files besides this one.

**NOTE**: however, this might still potentially impact downstream workflows, thus, should be noted in the release notes/changelog.

+ test_smallbaselineApp.py:
   - use the relative path to locate config file, instead of using mintpy module
   - remove the occurrence of MINTPY_HOME

**Reminders**

- [x] Pass Circle CI / local test (green)

### DIFF
--- a/mintpy/__init__.py
+++ b/mintpy/__init__.py
@@ -7,10 +7,11 @@ from mintpy.version import release_version, logo
 __version__ = release_version
 __logo__ = logo
 
-# check environment variable
-mintpy_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-sys.path.insert(1, mintpy_path)
-sys.path.insert(1, os.path.join(mintpy_path, 'defaults'))
-sys.path.insert(1, os.path.join(mintpy_path, 'objects'))
-sys.path.insert(1, os.path.join(mintpy_path, 'simulation'))
-sys.path.insert(1, os.path.join(mintpy_path, 'utils'))
+## check environment variable
+#mintpy_path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+#sys.path.insert(1, mintpy_path)
+#sys.path.insert(1, os.path.join(mintpy_path, 'defaults'))
+#sys.path.insert(1, os.path.join(mintpy_path, 'objects'))
+#sys.path.insert(1, os.path.join(mintpy_path, 'simulation'))
+#sys.path.insert(1, os.path.join(mintpy_path, 'utils'))
+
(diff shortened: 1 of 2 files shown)


test/test_smallbaselineApp.py
```diff
@@ -34,7 +32,9 @@
     'https://zenodo.org/record/3952917/files/KujuAlosAT422F650.tar.xz',
 ]
 
-PROJ_NAME_LIST = [os.path.basename(url).split('.tar.xz')[0] for url in URL_LIST]
+PROJ_NAMES = [os.path.basename(url).split('.tar.xz')[0] for url in URL_LIST]
+TEMPLATE_FILES = [os.path.join(os.path.dirname(__file__), 'configs/{}.txt'.format(proj_name))
+                  for proj_name in PROJ_NAMES]
```
jhkennedy, line 37:
```suggestion
TEMPLATE_FILES = [os.path.join(os.path.dirname(os.path.abspath(__file__)), 'configs/{}.txt'.format(proj_name))
                  for proj_name in PROJ_NAMES]
```

I checked out your branch and tried a
```sh
python test/test_smallbaselineApp.py
```
which failed with
```
FileNotFoundError: test/configs/FernandinaSenDT128.txt
```
Adding the above suggested change, however, made it work.

Suprisingly, in this case `__file__` isn't an absolute path, which it needs to be since [`smallBaselineApp.py` is called](https://github.com/insarlab/MintPy/pull/653/files#diff-96f2c9926af3120f234f57b879cf7c88579654d960e4cbf9d42d0fb688a827c2R148) from [the `work_dir`](https://github.com/insarlab/MintPy/pull/653/files#diff-96f2c9926af3120f234f57b879cf7c88579654d960e4cbf9d42d0fb688a827c2R128-R129)

---
You could also make it more compact with:
```python
from pathlib import Path
TEMPLATE_FILES = [Path(__file__).resolve().parent / 'configs' / f'{proj_name}.txt'
                  for proj_name in PROJ_NAMES]
```

jhkennedy (changes requested):
Looks great! Just one real change I'd suggest (`TEMPLATE_FILES`) and another structure suggestion (namespace) you can feel free to take or leave.

mintpy/__init__.py
```diff
@@ -14,9 +14,3 @@
 sys.path.insert(1, os.path.join(mintpy_path, 'objects'))
 sys.path.insert(1, os.path.join(mintpy_path, 'simulation'))
 sys.path.insert(1, os.path.join(mintpy_path, 'utils'))
```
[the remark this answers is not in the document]
yunjunz: @jhkennedy I like your suggestion and have noted it in the commit msg title/content. This should be enough to remind me to note it in the next release note.

test/test_smallbaselineApp.py
yunjunz: Thank you for the suggestion, I have added it.

jhkennedy (approved):
Looks great!

Issue #497, opened 2020-12-28
installation process via conda and setup.py


pbrotoisworo:
Hello everyone. May I recommend the following revisions to the installation process below. I've been testing it on my personal download of MintPy and it can run the MintPy analysis just fine. I'm hoping this will work in other operating systems and so that you don't have to make separate instructions for Linux/MacOS/Windows.

My testing environment:

- Windows 8.1/Windows 10
- Python 3.8 Anaconda

**Description of the desired feature**

Update MintPy installation process so that it is easier to install.

**Describe the solution you'd like**

Here are some of the revisions I did in my testing environment.

1.  Create a `setup.py` for [pyaps3](https://github.com/yunjunz/pyaps3)
I've been using a simple setup script as seen in the codeblock below and it works for me. I'm able to import `pyaps3` as a normal Python library. I `cd` into the pyaps directory and use `python setup.py install`.

```python
from setuptools import setup, find_packages

setup(
    name='pyaps3',
    version='1.0',
    author='Angelique Benoit',
    packages=find_packages(include=['pyaps3', 'pyaps3.*']),
    package_data={'': ['*.cfg', 'LICENSE']},
    include_package_data=True
)
```

2. Remove `openmp` from MintPy `setup.py` file.

`openmp` does not exist in the pip library. It exists in the conda library [here](https://anaconda.org/conda-forge/openmp). In my experience you can run `python setup.py install` if you remove `openmp` and you can run and import the MintPy software without any issue. However, I noticed some errors pop up due to improper formatting in `lookup_geo2radar.py` as seen in the image below.

![image](https://user-images.githubusercontent.com/32087633/103189263-c8a7be80-4906-11eb-9dc3-ca8bd62ed516.png)

**Notes on Config Files**:
Installing in the standard python path will make the config file harder to access. We can access the location of the CFG file by doing this:

```python
    import pyaps3
    import os

    pyaps3_cfg_dir = os.path.dirname(pyaps3.__file__)
    pyaps3_cfg_path = os.path.join(pyaps3_cfg_dir, 'model.cfg')
```

**My installation process**:
After those revisions my final installation process is as follows:

1.  Create virtual environment using provided `conda_env.yml` file
2. Run `python setup.py install` for MintPy
3. Run `python setup.py install` for PyAPS3

I did not need to setup any environment variables.

**Are you willing to help implement and maintain this feature?** Yes/No
Yes.

I'm not sure if this method works on MacOS and Linux operating systems. May someone try?

[...]

PR #659, opened 2021-09-09
setup: add all executable modules as `entry_points/console_scripts`

jhkennedy:

This  add all currently executable modules as `console_scripts` `entry_points` in `setup.py`, addressing Task 3 in #652 and (I think) finished off #497. [This comment](https://github.com/insarlab/MintPy/issues/652#issuecomment-914038914) gives a detailed explination of how entrypoints work, and how it changes develoment workflows. 

*I have **not** extensively tested these scripts beyond making sure arguments are passed through appropriately and all can be called with `--help`. 

**Note:** I've not added [`tropo_pyaps.py`](https://github.com/insarlab/MintPy/compare/main...jhkennedy:entrypints?expand=1#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R122) as I believe it's a legacy script that's been superseded by `tropo_pyaps3.py`. 

**Other thoughts:**
* It might be worth rearranging the `setup.py` file (e.g., move the `packages` and `entry_points` arguments to the bottom)

**Reminders**


jhkennedy:
Oh, I also have not addressed any installation, etc. type documentation -- that could be addressed here, or in conjunction with documentation about conda-forge after this has been vetted more.

mintpy/save_gbis.py
```diff
@@ -62,9 +62,9 @@ def create_parser():
 
 
 def cmd_line_parse(iargs=None):
-    print('{} {}'.format(os.path.basename(__file__), ' '.join(iargs)))
     parser = create_parser()
     inps = parser.parse_args(args=iargs)
+    print('{} {}'.format(os.path.basename(__file__), ' '.join(iargs)))
```
jhkennedy, line 67: Moving this below `parse_args` so that help and usage can be printed (and exit) gracefully if `iargs` is `None`

mintpy/multi_transect.py
```diff
@@ -393,7 +393,7 @@ def read_fault_coords(Fault_coord_file, Dp):
 
 
 #####################################################################
-def main(argv):
+def main(argv=None):
```
jhkennedy, line 396: Entrypoints get called with no arguments, so this needs to be optional. Seems to work fine with just this change.

mintpy/add_attribute.py
```diff
@@ -74,7 +74,7 @@ def update_file_attribute(fname, atr_new):
     return fname
 
 
-def main(argv):
+def main(argv=None):
```
jhkennedy, line 77: Entrypoints get called with no arguments, so this needs to be optional. Seems to work fine with just this change.

mintpy/tsview.py
```diff
@@ -711,7 +711,7 @@ def __init__(self, cmd=None, iargs=None):
         if cmd:
             iargs = cmd.split()[1:]
         self.cmd = cmd
-        self.iargs = iargs
+        self.iargs = cmd_line_parse(iargs)
         # print command line
         cmd = '{} '.format(os.path.basename(__file__))
         cmd += ' '.join(iargs)
```
jhkennedy, line 717: moving parsing to `__init__` (which I think also makes more sense) so that help and usage can be printed (and exit) gracefully if `iargs` is `None`

mintpy/add_attribute.py
yunjunz: LGTM.

mintpy/multi_transect.py
yunjunz: LGTM. `add_attribute.py` and `multi_transect.py` seem to be the only two scripts having this issue. Thanks!

mintpy/tsview.py
yunjunz: `cmd_line_parse()` returns a Namespace object, while `self.iargs` is a list of strings, so I don't think this modification works.

mintpy/tsview.py
```diff
@@ -736,8 +736,7 @@ def __init__(self, cmd=None, iargs=None):
 
 
     def configure(self):
-        inps = cmd_line_parse(self.iargs)
-        inps, self.atr = read_init_info(inps)
+        inps, self.atr = read_init_info(self.iargs)
         # copy inps to self object
         for key, value in inps.__dict__.items():
             setattr(self, key, value)
```
yunjunz, line 742: The `inps` Namespace object is only generated here to be saved as member variables in the `self` object. Thus, it seems simpler to me to keep the parsing and saving next to each other.

>  so that help and usage can be printed (and exit) gracefully if iargs is None

I don't think I get what do you mean.

setup.py
```diff
@@ -57,7 +57,75 @@ def do_setup():
...
+                'subset.py = mintpy.subset:main',
+                'temporal_average.py = mintpy.temporal_average:main',
+                'temporal_derivative.py = mintpy.temporal_derivative:main',
+                'temporal_filter.py = mintpy.temporal_filter:main',
+                'timeseries2velocity.py = mintpy.timeseries2velocity:main',
+                'timeseries_rms.py = mintpy.timeseries_rms:main',
+                'tropo_gacos.py = mintpy.tropo_gacos:main',
+                'tropo_phase_elevation.py = mintpy.tropo_phase_elevation:main',
+                'tropo_pyaps3.py = mintpy.tropo_pyaps3:main',
+                # 'tropo_pyaps.py = mintpy.tropo_pyaps:main',
```
yunjunz, line 122: Let's keep both as `tropo_pyaps.py` is not completely replaced by `tropo_pyaps3.py` yet.

yunjunz:
> It might be worth rearranging the setup.py file (e.g., move the packages and entry_points arguments to the bottom)

Sounds good to me, please do it as you see fit.

yunjunz:
> Oh, I also have not addressed any installation, etc. type documentation -- that could be addressed here, or in conjunction with documentation about conda-forge after this has been vetted more.

Yep, there will be big (good) changes in the installation. I would prefer to do it in another PR when conda-forge is ready.

mintpy/tsview.py
jhkennedy: Addressed in comment above

mintpy/tsview.py
jhkennedy: If `iargs` is `None`, because [`main` got called with `iargs=None` ](https://github.com/jhkennedy/MintPy/blob/entrypints/mintpy/tsview.py#L1025-L1026), then without this change, `timeseriesViewer.__init__()` will crash out when [trying to form the `cmd` string](https://github.com/jhkennedy/MintPy/blob/entrypints/mintpy/tsview.py#L717) becuase:
```python
>>> iargs = None
>>> ' '.join(iargs)
  File "<stdin>", line 1, in <module>
TypeError: can only join an iterable
>>> 
```
So either you need a:
```python
if iargs is not None:
    cmd += ' '.join(iargs)
```

or to handle the parsing before this point. The benefit of parsing as early as possible, is that you'll exit *fast* (before other errors could happen) for `--help`/missing/invalid arguments. 

it doesn't look like `self.iargs` is used anywhere except in the later configure step, and so the change below accommodates the change to a `Namespace` object. If it's used/access elsewhere, the `if` block  might be more appropriate.

setup.py
jhkennedy: `tropo_pyaps.py` will fail with a:
```sh
$ tropo_pyaps.py
  File "/home/jhkennedy/Documents/Code/contributing/MintPy/mintpy/tropo_pyaps.py", line 14, in <module>
    import pyaps as pa
ModuleNotFoundError: No module named 'pyaps'
```
due to [this import](https://github.com/jhkennedy/MintPy/blob/entrypints/mintpy/tropo_pyaps.py#L13-L16) of the older (Python 2?) `pyaps` package` which isn't listed as a dependency or install in the environments.

Could it be converted to `import pyaps3 as pa`? Otherwise, I don't think it can/should be listed as an entrypoint.

yunjunz: `pyaps` package was distributed under Caltech/Earthdef. The website went down recently. I suppose there are still users (including me occasionally) using that via `tropo_pyaps.py`.

mintpy/tsview.py
yunjunz: Thanks for the explanation. Your `if` block looks good to me. I saved `self.iargs` specifically to check input options later on in other scripts/notebooks (should have said this earlier; view.py has the same setup as well).

setup.py
jhkennedy: Alright, I'll include it, but it'd probably be worth improving the [import error message](https://github.com/jhkennedy/MintPy/blob/entrypints/mintpy/tropo_pyaps.py#L16) to explain what pyaps is and where to get it.

mintpy/tsview.py
jhkennedy: Switched to the `if` block!

jhkennedy:
> > It might be worth rearranging the `setup.py` file (e.g., move the packages and entry_points arguments to the bottom)

> Sounds good to me, please do it as you see fit.

Done!

yunjunz (approved):
Looks all good to me. I will do more testings on the weekend before merging it. Thank you @jhkennedy!

yunjunz:
I installed and tested mintpy as below on macOS. Everything looks good to me.

```bash
# download
cd ~/tools
git clone https://github.com/insarlab/MintPy.git
# install
conda create --name test
conda activate test
conda install --file tools/MintPy/docs/requirements.txt pysolid pyaps3 gdal">=3"
cd ~/tools/MintPy
python setup.py install
# test
python ~/tools/MintPy/tests/test_smallbaselineApp.py
```

jhkennedy:
@yunjunz Just a note -- `python -m pip install` should be preferred over `python setup.py install` as `setup.py` invokes the old and depreciated egg protocol.

yunjunz:
Good to know, thanks.

PR #666, opened 2021-09-18
Plotting flexibility and support for GNSS component plotting

sssangha:

I have added in a series of interrelated, dependent changes concerning plotting flexibility of user-defined bounds, and support for plotting of vertical and horizontal GNSS components.

1. Allow users to plot data in expanded bounds. Originally the code superseded user-specified bounds in such cases in favor of tight bounds about the track. This flexibility is crucial in my experience as it makes it much easier to compile and mosaic figures of large study areas involving multiple tracks.

2. Mask DEM background where it falls outside of InSAR coverage or is coincident with masked/nodata values.

3. Suppress plotting of GPS stations that fall outside of InSAR coverage or are coincident with masked/nodata values. Again, this flexibility is a welcome addition for compiling figures. Perhaps we can make this a user-specified option (i.e. by default plot empty symbols where there are no valid InSAR pixels), but leave alone my other commit whereby such stations are not passed into the output CSV.

4. Introduce support for plotting vertical or horizontal GPS components if the user is plotting vertical or horizontal InSAR velocity components, respectively.

5. Update list of supported units to also support American convention (e.g. inches, feet, miles).

Working example before code allowed for bounds expanded beyond the range of the input dataset, and masked out GNSS stations and DEM pixels coincident with nodata values:
test2_velocity_TECTscale_GPS

After my introduced changes:
test_OG

### DIFF
--- a/mintpy/objects/coord.py
+++ b/mintpy/objects/coord.py
@@ -450,6 +450,26 @@ class coordinate:
             msg += '\tsubset range in (W, N, E, S): {}\n'.format(self.box_pixel2geo(pixel_box))
             raise ValueError(msg)
 
+        # Check Y/Azimuth/Latitude subset range
+        if sub_y[0] < 0:
+            sub_y[0] = 0
+            if print_msg:
+                print('WARNING: input y < min (0)! Set it to min.')
+        if sub_y[1] > length:
+            sub_y[1] = length
+            if print_msg:
+                print('WARNING: input y > max ({})! Set it to max.'.format(length))
+
+        # Check X/Range/Longitude subset range
+        if sub_x[0] < 0:
+            sub_x[0] = 0
+            if print_msg:
+                print('WARNING: input x < min (0)! Set it to min.')
+        if sub_x[1] > width:
+            sub_x[1] = width
+            if print_msg:
+                print('WARNING: input x > max ({})! Set it to max.'.format(width))
+
         out_box = (sub_x[0], sub_y[0], sub_x[1], sub_y[1])
         return out_box
 
--- a/mintpy/objects/coord.py
+++ b/mintpy/objects/coord.py
@@ -450,26 +450,6 @@ class coordinate:
             msg += '\tsubset range in (W, N, E, S): {}\n'.format(self.box_pixel2geo(pixel_box))
             raise ValueError(msg)
 
-        # Check Y/Azimuth/Latitude subset range
-        if sub_y[0] < 0:
-            sub_y[0] = 0
-            if print_msg:
-                print('WARNING: input y < min (0)! Set it to min.')
-        if sub_y[1] > length:
-            sub_y[1] = length
-            if print_msg:
-                print('WARNING: input y > max ({})! Set it to max.'.format(length))
-
-        # Check X/Range/Longitude subset range
-        if sub_x[0] < 0:
-            sub_x[0] = 0
-            if print_msg:
-                print('WARNING: input x < min (0)! Set it to min.')
-        if sub_x[1] > width:
-            sub_x[1] = width
-            if print_msg:
-                print('WARNING: input x > max ({})! Set it to max.'.format(width))
-
         out_box = (sub_x[0], sub_y[0], sub_x[1], sub_y[1])
         return out_box
 
--- a/mintpy/objects/gps.py
+++ b/mintpy/objects/gps.py
@@ -16,7 +16,7 @@ from pyproj import Geod
 from urllib.request import urlretrieve
 
 from mintpy.objects.coord import coordinate
-from mintpy.utils import ptime, time_func, readfile, utils1 as ut, utils as ut0
+from mintpy.utils import ptime, time_func, readfile, utils1 as ut
 
 
 unr_site_list_file = 'http://geodesy.unr.edu/NGLStationPages/DataHoldings.txt'
@@ -164,7 +164,7 @@ def get_gps_los_obs(insar_file, site_names, start_date, end_date, msk,
             geom_obj = meta
             vprint('use incidence / azimuth angle from metadata')
         cropped_metadata = get_plot_extent(metadata, msk, geo_box)
-        coord = ut0.coordinate(cropped_metadata)
+        coord = coordinate(cropped_metadata)
 
         # loop for calculation
         prog_bar = ptime.progressBar(maxValue=num_site, print_msg=print_msg)
@@ -209,8 +209,6 @@ def get_gps_los_obs(insar_file, site_names, start_date, end_date, msk,
             fcw.writerows(data_list_tocsv)
 
     return site_obs
-
-
 
 
 #################################### Beginning of GPS-GSI utility functions ########################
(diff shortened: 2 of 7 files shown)


yunjunz:
Thank you @sssangha for contributing. All your 5 new features sound good to me in general. I have a few specific suggestions below that I hope you could address before considering merging the PR. Please do not let my strict review discourage you, view.py is one of the most used scripts, I just want to make sure it does not break.

1. both circle ci and codacy flagged a few issues, could you address them?
2. For DEM masking, could we add an option for it, e.g. `--mask-dem` with no masking as default?
3. For GPS masking, same as DEM.

mintpy/objects/gps.py
```diff
@@ -16,7 +16,7 @@
 from urllib.request import urlretrieve
 
 from mintpy.objects.coord import coordinate
-from mintpy.utils import ptime, time_func, readfile, utils1 as ut
+from mintpy.utils import ptime, time_func, readfile, utils1 as ut, utils as ut0
```
yunjunz, line 19: `utils1` already has everything from `utils0`, so this additional import is not needed.
https://github.com/insarlab/MintPy/blob/670c6a9f9657f883ad6563d4f4e66bf6f8c7aac3/mintpy/utils/utils1.py#L21

mintpy/objects/coord.py
```diff
@@ -450,26 +450,6 @@ def check_box_within_data_coverage(self, pixel_box, print_msg=True):
...
-        if sub_x[0] < 0:
-            sub_x[0] = 0
-            if print_msg:
-                print('WARNING: input x < min (0)! Set it to min.')
-        if sub_x[1] > width:
-            sub_x[1] = width
-            if print_msg:
-                print('WARNING: input x > max ({})! Set it to max.'.format(width))
-
         out_box = (sub_x[0], sub_y[0], sub_x[1], sub_y[1])
```
yunjunz, line 453: `check_box_within_data_coverage()` is used in other places besides view.py. For the purpose described in this PR, we should just not calling this function inside view.py.

mintpy/utils/arg_group.py
```diff
@@ -165,16 +165,24 @@ def add_gps_argument(parser):
...
-                     help='start date of GPS data, default is date of the last SAR acquisiton')
+                     help='start date of GPS data, default is date of the last SAR acquisition')
+    gps.add_argument('--azimuth', '--az', dest='az_angle', type=float, default=0.,
+                     help='azimuth angle in degree (clockwise) of the direction of the horizontal movement\n' +
+                             'default is 0., assuming no projection needed and magnitude of horizontal vectors passed.\n' +
+                             'i.e. azimuth angle of strike-slip fault\n\n' +
+                             'Note:\n' +
+                             'a. Near north direction can not be well resolved due to the lack of\n' +
+                             '   diversity in viewing geometry. Check exact dilution of precision for \n' +
+                             '   each component in Wright et al., 2004, GRL')
```
yunjunz, line 185: I made sure recently that all azimuth angles in mintpy are [measured from the north with anti-clockwise direction as positive in the unit of degrees](https://github.com/insarlab/MintPy/blob/670c6a9f9657f883ad6563d4f4e66bf6f8c7aac3/mintpy/utils/utils0.py#L397). Could you adjust this for consistency?

sssangha:
> Thank you @sssangha for contributing (your data is beautiful!). All your 5 new features sound good to me in general. I have a few specific suggestions below that I hope you could address before considering merging the PR. Please do not let my strict review discourage you, view.py is one of the most used scripts, I just want to make sure it does not break.

Thanks @yunjunz! And yes, I agree. We need to be careful with commits especially to view.py. Honestly, I'm always a bit nervous when committing to this script in particular in mintpy, and 'extractProduct.py' in ARIA-tools!

Anyways, I've integrated all of your suggestions and the tests seem to run fine locally, but it looks like I may need to sort out some circleci issues. I will confirm here once that all seems to be addressed.

mintpy/objects/gps.py
sssangha: 'coordinate' module used directly now.

mintpy/objects/coord.py
sssangha: Done, this call is removed from view.py

mintpy/utils/arg_group.py
sssangha: Accordingly adjusted in the code and parser

sssangha:
@yunjunz it looks like I sorted out the codacy and circleCI issues. Please let me know if you have any other questions/concerns. Thanks again

mintpy/objects/coord.py
yunjunz: Let's also do not modify this function also so that other scripts won't be affected.

sssangha: @yunjunz ok, went ahead and tweaked this accordingly. Let me know what you think.

sssangha:
@yunjunz I believe I have addressed all of your initial comments/concerns with my latest commits, so please let me know if any further action or review is required before we review. Thanks.

yunjunz:
Thank you @sssangha. I will check your PR again tonight.

yunjunz:
Hi @sssangha, I find it hard to see the flow of changes (and potential bugs), because the 1st feature requires complicated modifications (due to the complex script itself), while the rest 4 features are straightforward. Would you please consider separating this PR into two PRs: one for the extended bounds and one for the rest?

sssangha:
@yunjunz sure, I will take care of this later Tuesday.

Would it be possible to let me know though or help resolve potential conflicts if you plan on committing anything involved with the scripts in this PR, in case we both attempt commits today? Thanks

yunjunz:
🙌🏼 No planned commit from me.

yunjunz:
@sssangha I forgot to mention that `subset.py` supports extended bounds. You might want to check the code there also: https://github.com/insarlab/MintPy/blob/670c6a9f9657f883ad6563d4f4e66bf6f8c7aac3/mintpy/subset.py#L377

sssangha:
> @sssangha I forgot to mention that `subset.py` supports extended bounds. You might want to check the code there also. It might be a good idea to have related new functions sit there. The existing functions are not well commented though, please feel free to modify them as needed.
> 
> https://github.com/insarlab/MintPy/blob/670c6a9f9657f883ad6563d4f4e66bf6f8c7aac3/mintpy/subset.py#L377

I was thinking about this, but the problem is that the relevant function generates a physical output file, which would lead to a great deal of overhead and inflexibility (e.g. imagine physical outputs for an entire TS stack and large DEM just for plotting purposes). 

The proposed function I've written is pretty short and would adjust data bounds on the fly for plotting, but I could add the function to this script for organization purposes. What do you think?

yunjunz:
I agree with you, the current form of sub-functions in `subset.py` is not good enough for plotting purposes. 

> The proposed function I've written is pretty short and would adjust data bounds on the fly for plotting, but I could add the function to this script for organization purposes. What do you think?

I like this idea.

yunjunz:
Hi @sssangha, sorry for the late following up. With the other functionalities merged, shall we rebase this PR (or issue a new PR if you prefer) for the expanded bounds in view.py?

It will be great to have an example usage with a short explanation in the [view.py#L45](https://github.com/insarlab/MintPy/blob/7014b0b328da5819374dae01e8defd96b6d42646/mintpy/view.py#L45) as well.

PR #670, opened 2021-09-23
plot: --mask-gps/dem, --gps-comp horz/vert and American display unit

sssangha:

Following discussion in PR #666 , I plan to push two separate and successive PRs, starting with this one which introduces the following changes/options:

- [x] 1. Option to Mask DEM background, where it falls outside of InSAR coverage or is coincident with masked/nodata values (`--mask-dem`).

- [x] 2. Option to mask GPS stations that fall outside of InSAR coverage or are coincident with masked/nodata values (`--mask-gps`).

- [x] 3. Introduce support for plotting vertical or horizontal GPS components if the user is plotting vertical or horizontal InSAR velocity components, respectively, via `--gps-comp horz/vert` and `--horz-az` options.

- [x] 4. Update list of supported units to also support American convention (e.g. inches, feet, miles).

@yunjunz instead of issuing the other PR I want right now -- which supports plotting of expanding bounds -- I think it would be easier to first have this PR merged. Plus perhaps for reference we should keep PR #666 open for now, until both proposed PRs are merged? What do you think?

**Reminders**


yunjunz:
> I think it would be easier to first have this PR merged. Plus perhaps for reference we should keep PR #666 open for now, until both proposed PRs are merged? What do you think?

Having this PR merged first sounds good @sssangha. I will take a closer look at the PR tonight.

mintpy/utils/readfile.py
```diff
@@ -772,6 +772,10 @@ def read_attribute(fname, datasetName=None, metafile_ext=None):
             k = list(set(g1_list) & set(py2_mintpy_stack_files))[0]
         elif 'FILE_TYPE' in atr:
             k = atr['FILE_TYPE']
+            # catech exception for horizontal and vertical decomposition
+            if (k == 'None') and ('SenD' in d1_list or 'horizontal' \
+                in d1_list or 'vertical' in d1_list):
+                k = 'SenD'
```
yunjunz, line 778: I guess you add this because of the improper FILE_TYPE in the `asc_desc2horz_vert.py` output? If so, we should update there instead, e.g. from 

https://github.com/insarlab/MintPy/blob/6c7ad9ec07462022f021886b92f02164bf842ef6/mintpy/asc_desc2horz_vert.py#L267-L268

to 

```python
if dsname and atr['FILE_TYPE'] in ['ifgramStack', 'timeseries', 'HDFEOS']:
        atr['FILE_TYPE'] = 'displacement'
```

Then no change is needed here in `readfile.py`.

mintpy/utils/plot.py
```diff
@@ -1635,10 +1658,18 @@ def draw_scalebar(ax, geo_box, unit='degrees', loc=[0.2, 0.2, 0.1], labelpad=0.0
     ax.plot([lon1, lon1], [lat_c, lat_c + 0.1*length_disp], color=color)
 
     ## plot scale bar label
-    unit = 'm'
-    if length_meter >= 1000.0:
-        unit = 'km'
-        length_meter *= 0.001
+    if unit.split('/')[0]=='in' or unit.split('/')[0]=='ft' or unit.split('/')[0]=='mi':
```
yunjunz, line 1661:
```suggestion
    if unit.split('/')[0] in ['in', 'ft', 'mi']:
```

mintpy/utils/plot.py
```diff
@@ -1154,8 +1160,16 @@ def plot_gps(ax, SNWE, inps, metadata=dict(), print_msg=True):
 
         # plot
         for lat, lon, obs in zip(site_lats, site_lons, site_obs):
```
yunjunz, line 1162:
```suggestion
        # Mask GNSS stations if outside of data range (resulting in NaN values)
        flag = ~np.isnan(site_obs) if inps.gps_mask else np.ones(site_obs.shape, dtype=np.bool_)
        for lat, lon, obs in zip(site_lats[flag], site_lons[flag], site_obs[flag]):
```

mintpy/utils/readfile.py
sssangha: Yeah I agree, this would be much cleaner and the more generic designation (i.e. displacement vs. SenD) makes sense to me

yunjunz:
The changes related to masking in `plot.plot_gps()` and `gps.get_gps_los_obs()` seem overcomplicated. We should apply the checking you have for `NaN data value` and `zero in mask value` right after searching the list of GNSS sites below. Then everything afterward is consistent between plot and csv and same as before.

https://github.com/insarlab/MintPy/blob/6c7ad9ec07462022f021886b92f02164bf842ef6/mintpy/utils/plot.py#L1108

mintpy/objects/gps.py
```diff
@@ -168,37 +172,43 @@ def get_gps_los_obs(insar_file, site_names, start_date, end_date,
             # calculate gps data value
             obj = GPS(site_name)
 
-            if obs_type == 'velocity':
+            if obs_type == 'velocity' or obs_type == 'SenD':
                 vel, dis_ts = obj.get_gps_los_velocity(geom_obj,
                                                        start_date=start_date,
                                                        end_date=end_date,
-                                                       gps_comp=gps_comp)
+                                                       ref_site=ref_site,
```
yunjunz, line 179: I did not add referencing on purpose before, so that data in the CSV file could be more generic. Note that the referencing is applied as below right after this.

https://github.com/insarlab/MintPy/blob/6c7ad9ec07462022f021886b92f02164bf842ef6/mintpy/utils/plot.py#L1140

mintpy/view.py
```diff
@@ -1572,12 +1580,17 @@ def plot(self):
...
                 vprint('masking data')
                 data = np.ma.masked_where(self.msk == 0., data)
+            if self.msk is None:
+                self.msk = np.ones(data.shape, dtype=np.int8)
+            if self.zero_mask:
+                vprint('masking pixels with zero value')
+                self.msk = np.ma.masked_array(self.msk, mask=np.isnan(data))
+                np.ma.set_fill_value(self.msk, 0)
+                self.msk=self.msk.filled()
+                data = np.ma.masked_where(self.msk == 0., data)
```
yunjunz, line 1593: I don't get the reason behind this change, could you describe?

yunjunz:
Thank you for the PR @sssangha, it's much more clear. Feature 4 looks good to me. Could you address the suggestions I made on features 1 and 2. After that, I will check feature 3, as it has related code to feature 2.

mintpy/view.py
sssangha: The rationale was to update the mask to reflect 0-data values in the input data, if specified. Makes sense?

mintpy/objects/gps.py
sssangha: Do you think perhaps it may be a better for the CSV to reflect the plot, especially since it may make it more convenient for the user to take the CSV and use it for their work through other programs/interfaces (e.g. RAiDER, QGIS, etc.).
I understand your rationale to keep it generic and it should be simple for the user to manually tweak the CSV to reflect the specified reference station, but then again it makes it much more convenient to take these outputs directly as part of a broader, automated routine or so. What do you think?

sssangha:
@yunjunz sounds good, I'm preparing a push soon where I integrate the suggestions. I will confirm with you once the checks pass for you to review. Thanks again

mintpy/objects/gps.py
yunjunz: Good point. Let keep `ref_site` then. 

One detail: I would recommend, inside `get_gps_los_obs()`, using a simple referencing as we have in `plot.plot_gps()`, instead of passing ref_site to `read_gps_los_velocity/displacement()` for faster speed and more robust, because `read_gps_los_displacement()` is grabbing the common acquisitions between GPS pairs. It's fine to compare time-series, which is my original purpose, but can be a potential issue for displacement/velocity when the two sites has sparse and uncommon dates.

mintpy/view.py
yunjunz: I meant: is the original code not working? If so, why?

sssangha: Originally, the code was not working as intended -- inline with my commits -- as the mask in memory was not updated to reflect the zero pixels in the input data. My commits access this mask in order to mask the DEM and GPS stations, if specified.

mintpy/objects/gps.py
sssangha: Ok implemented as suggested, I hope it's cleaner/simpler! However, not that I left alone the sections of the `read_gps_los_velocity/displacement()` functions which handle the ref station. I don't specify the reference stations in the respective function call, but I just left those functions alone.

sssangha:
@yunjunz ok, looks like I successfully sorted out some codacy snags! Let me know what you think of my updates.

mintpy/objects/gps.py
```diff
@@ -101,8 +101,9 @@ def get_baseline_change(dates1, pos_x1, pos_y1, pos_z1,
     return dates, bases
 
 
-def get_gps_los_obs(insar_file, site_names, start_date, end_date,
-                    gps_comp='enu2los', print_msg=True, redo=False):
+def get_gps_los_obs(insar_file, site_names, start_date, end_date, msk,
+                    geo_box, metadata, gps_comp='enu2los', print_msg=True, redo=False,
+                    ref_site=None, az_angle=0.):
```
yunjunz, line 106: `msk, geo_box and metadata` should be gone.

mintpy/objects/gps.py
```diff
@@ -168,11 +169,12 @@ def get_gps_los_obs(insar_file, site_names, start_date, end_date,
             # calculate gps data value
             obj = GPS(site_name)
 
-            if obs_type == 'velocity':
+            if obs_type == 'velocity' or obs_type == 'displacement':
```
yunjunz, line 172: `displacement` should be together with `timeseries`.

mintpy/objects/gps.py
yunjunz: that sounds great.

mintpy/objects/gps.py
yunjunz: Or how about we calculate/save the velocity all the time by getting rid of the `if elif` and use `obj.get_gps_los_velocity()` only? We would need to adjust `site_obs` line below for the obs_type check.

mintpy/utils/plot.py
```diff
@@ -917,7 +917,7 @@ def plot_coherence_matrix(ax, date12List, cohList, date12List_drop=[], p_dict={}
     return ax, coh_mat, im
 
 
-def read_dem(dem_file, pix_box=None, geo_box=None, print_msg=True):
+def read_dem(dem_file, pix_box=None, geo_box=None, print_msg=True, multilook_num=1):
```
yunjunz, line 920: `multilook_num` is not used here. And note that the DEM here could have a higher resolution than the data, so it's fine to not multilook.

mintpy/utils/arg_group.py
```diff
@@ -163,18 +165,29 @@ def add_gps_argument(parser):
...
     gps.add_argument('--ref-gps', dest='ref_gps_site', type=str, help='Reference GPS site')
 
     gps.add_argument('--gps-start-date', dest='gps_start_date', type=str, metavar='YYYYMMDD',
-                     help='start date of GPS data, default is date of the 1st SAR acquisiton')
+                     help='start date of GPS data, default is date of the 1st SAR acquisition')
     gps.add_argument('--gps-end-date', dest='gps_end_date', type=str, metavar='YYYYMMDD',
-                     help='start date of GPS data, default is date of the last SAR acquisiton')
+                     help='start date of GPS data, default is date of the last SAR acquisition')
+    gps.add_argument('--azimuth', '--az', dest='az_angle', type=float, default=0.,
+                     help='azimuth angle of the horizontal movement in degrees\n'
```
yunjunz, line 183: We should use `None` as default because 0 deg az angle has valid meaning.

mintpy/view.py
yunjunz: not really. The previous version of PR broke the masking of pixels with zero data value. I have reverted it. I have saved the `self.msk` for the purpose of this PR, so it should still work for you.

mintpy/objects/gps.py
yunjunz: Now I remembered that: no-referencing is required to support spatially absolute comparison bwt. SAR and GPS. I have re-moved it to `plot.plot_gps()` again and add some comments for future reference.

mintpy/utils/plot.py
yunjunz: this func is used only for single plot, which does not use multilook_num. Revert this change.

mintpy/utils/arg_group.py
yunjunz: -90 is used as default instead, to be consistent with `asc_desc2horz_vert.py`. The option is also renamed to `--horz-az` to be not confused with az_angle of the LOS vector.

yunjunz:
This PR looks all good to me now. Thank you @sssangha for the many new features!

We spent quite some time together during this and previous PR to figure the flow of changes. They really should be broken down into separate PRs next time.

yunjunz (approved):

sssangha:
> This PR looks all good to me now. Thank you @sssangha for the many new features!
> 
> You and I spent quite some time during this and previous PR to figure the flow of changes. They really should be broken down into separate PRs (one PR for one new feature or behavior change) next time. You might want to do the same as much as possible for your own commits as well.

Yes, my apologies. There were a series of old changes I built off a ~1 yr old fork and I was really -- probably a bit too much so -- motivated to get them all in one shot.

With that in mind, I actually am preparing to push a simple PR to correct a bug I found related to one specific aspect of this PR! Will again update you when it's ready.

yunjunz:
No worries at all, and thank you again for the features. They do look good in the plotting!

Issue #123, opened 2019-07-16
regularly updated docker image


andretheronsa:
For convenience I have created a [docker](https://docs.docker.com/get-started/) image on Docker hub for Mintpy: ``https://hub.docker.com/r/andretheronsa/mintpy``
If this would be useful for the project, it can be added it as an installation option ?

It provides an easy way to have the configured MintPy environment ready.  
It is built on a base debian:stretch linux image using the installation instructions (setting up conda, environment, mintpy, pyaps, pykml etc).  

All you need is docker installed (on windows / linux / mac).
Pull the image from dockerhub with: ``docker pull andretheronsa/mintpy:latest``
Run an interactive shell session in the container with: ``docker run -it andretheronsa/mintpy:latest bash``
To map your data folder on host to the container add ``-v /path/to/data/dir:/home/work/`` to the docker run command.

The image is the first version and has not been tested much. It can be made smaller too.

yunjunz:
This is great @andretheronsa! Yes, do a pull request to the `installation.md` please.

I don't have experience on `docker` yet. Is this image built based on the latest version on `insarlab/master` branch?

hfattahi:
This is awesome and very useful. Maybe it should build the released version which is supposed to be stable. @dbekaert mentioned that ARIA folks where working on a docker image for Mintpy as well. David please feel free to add ARIA folks to the discussion if needed.

dbekaert:
The main idea is to hook the circle-ci to docker such it attempt to re-build the docker regularly. Copying in @pymonger. We have not started with this yet, so could be good to combine the efforts.

andretheronsa:
Great to see the excitement!

@yunjunz I have done a pull request for updated installation instructions.
The container was built yesterday - it pulls the latest version of insarlab/master from git during build time. So it will need to be rebuild with new releases.

@dbekaert  I do not have experience with Circle-ci but such a tool to rebuild the docker will be useful.

I have added the dockerfile in a root level docker directory to the pull request, I am not sure if this is the right place for it, but you guys can check out the file to see exactly how it creates the container.

Note that if you build a docker image yourself from the file it will copy over all content in the directory you build it from - that is why it is in it's own directory. The rest of Mintpy gets pulled into the container from github.

andretheronsa:
After testing I just want to note two errors that had to be fixed in the docker container.
1) I had to install the zip package (for save_kmz.py etc.).
[...]

PR #685, opened 2021-11-01
Dockerfile: rewrite using Micromamba Image

forrestfwilliams:
References issue #123 
A recent build can be found at `https://hub.docker.com/repository/docker/forrestwilliams/mintpy`

- Rewrites the MintPy dockerfile using the mambaorg/micromamba image
- Shorter build times and slightly smaller image!
- Forces Python=3.6 for faster image build
- Sets the user as micromamba, not root
- Optionally allows the image to start with a Jupyter Lab instance

**Reminders**

- [x] Fix #xxxx

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,29 +10,33 @@ LABEL \
     "Dockerfile Author"="Forrest Williams" \
     "Email"="forrestfwilliams@icloud.com"
 
-# Install git and vim
+# Install command line tools: git, vim and wget
 USER root
 RUN apt-get update && \
     apt-get install -y git vim wget && \
     rm -rf /var/lib/apt/lists/*
 
-# Clone MintPy and PyAPS repositories
+# Setup path / environment for MintPy
 USER micromamba
 WORKDIR /home/micromamba
 
+ARG MINTPY_HOME=/home/micromamba/tools/MintPy
+ARG PYAPS_HOME=/home/micromamba/tools/PyAPS
+
+ENV PATH=${MINTPY_HOME}/mintpy:/opt/conda/bin:${PATH}
+ENV PYTHONPATH=${MINTPY_HOME}:${PYAPS_HOME}
+
+# Download source code
 RUN mkdir /home/micromamba/tools && \
-    git clone https://github.com/insarlab/MintPy.git /home/micromamba/tools/MintPy && \
-    git clone https://github.com/yunjunz/PyAPS.git /home/micromamba/tools/PyAPS
+    git clone https://github.com/insarlab/MintPy.git ${MINTPY_HOME} && \
+    git clone https://github.com/yunjunz/PyAPS.git ${PYAPS_HOME}
 
-# Prepare environment for MintPy
-ENV PATH=/home/micromamba/tools/MintPy/mintpy:/opt/conda/bin:${PATH}
-ENV PYTHONPATH=/home/micromamba/tools/MintPy:/home/micromamba/tools/PyAPS
-
+# Install dependencies
 # # # Optionally add Jupyter Lab to environment file
-# # RUN echo "  - jupyterlab\n  - ipympl" >> /home/micromamba/tools/MintPy/docs/environment.yml
+# # RUN echo "  - jupyterlab\n  - ipympl" >> ${MINTPY_HOME}/docs/environment.yml
 
 # ADD mintpy.yml /tmp
-RUN micromamba install -y -n base -f /home/micromamba/tools/MintPy/docs/environment.yml python=3.6 && \
+RUN micromamba install -y -n base -f ${MINTPY_HOME}/docs/environment.yml python=3.6 && \
     micromamba clean --all --yes
 
 # # Have the container start with a Jupyter Lab instance
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 # Use mambaforge as the base image
 # Builds in ~9.5 min and is ~ 2.5 GB on a windows laptop
-FROM mambaorg/micromamba
+FROM mambaorg/micromamba:0.15.3
 
 # Label image
 LABEL \
@@ -17,6 +17,7 @@ ARG user=micromamba
 USER root
 RUN apt-get update && \
     apt-get install -y git vim && \
+    rm /var/lib/apt/lists/* && \
     mkdir /home/micromamba/tools && \
     git clone https://github.com/insarlab/MintPy.git /home/micromamba/tools/MintPy && \
     git clone https://github.com/yunjunz/PyAPS.git /home/micromamba/tools/PyAPS
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -27,7 +27,7 @@ ENV PATH=${MINTPY_HOME}/mintpy:/opt/conda/bin:${PATH}
 ENV PYTHONPATH=${MINTPY_HOME}:${PYAPS_HOME}
 
 # Download source code
-RUN mkdir /home/micromamba/tools && \
+RUN mkdir -p ${MINTPY_HOME} ${PYAPS_HOME} && \
     git clone https://github.com/insarlab/MintPy.git ${MINTPY_HOME} && \
     git clone https://github.com/yunjunz/PyAPS.git ${PYAPS_HOME}
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,7 +13,8 @@ LABEL \
 # Install command line tools: git, vim and wget
 USER root
 RUN apt-get update && \
-    apt-get install -y git vim wget && \
+    apt-get install -y --no-install-recommends git vim wget && \
+    apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
 # Setup path / environment for MintPy
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,14 +17,16 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 # Clone MintPy and PyAPS repositories
-USER $user
-RUN mkdir /home/$user/tools && \
-    git clone https://github.com/insarlab/MintPy.git /home/$user/tools/MintPy && \
-    git clone https://github.com/yunjunz/PyAPS.git /home/$user/tools/PyAPS
+USER micromamba
+WORKDIR /home/micromamba
+
+RUN mkdir /home/micromamba/tools && \
+    git clone https://github.com/insarlab/MintPy.git /home/micromamba/tools/MintPy && \
+    git clone https://github.com/yunjunz/PyAPS.git /home/micromamba/tools/PyAPS
 
 # Prepare environment for MintPy
-ARG MINTPY_HOME=/home/$user/tools/MintPy
-ARG PYAPS_HOME=/home/$user/tools/PyAPS
+ARG MINTPY_HOME=/home/micromamba/tools/MintPy
+ARG PYAPS_HOME=/home/micromamba/tools/PyAPS
 ARG PYTHON3DIR=/opt/conda
 
 ENV MINTPY_HOME=${MINTPY_HOME}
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,72 +1,47 @@
...
-    git \
-    python3 \
-    python3-pip \
-    wget \ 
-    vim \
-    zip \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+# Set user name (the micromamba image comes with the micromamba user)
+ARG user=micromamba
```
jhkennedy, line 14: @forrestfwilliams what's your intention with this argument? 

Since this has to be set at build time, anyone pulling the container will get whatever was used. It probably makes the most sense to just leave it as the user that comes with the micromamba image. 

Otherwise, if you expect users to build the container they are likely going to want to specify both the UID and GID as they get mapped directly from host to container, so you'd need add a user with those ids or change the micromamba UID/GID.

docker/Dockerfile
```diff
@@ -1,72 +1,47 @@
...
-# Change default shell from dash to bash permanently
-SHELL ["/bin/bash", "-c"]
+# Clone MintPy and PyAPS repositories
+USER root
+RUN apt-get update && \
+    apt-get install -y git vim && \
+    rm /var/lib/apt/lists/* && \
+    mkdir /home/micromamba/tools && \
+    git clone https://github.com/insarlab/MintPy.git /home/micromamba/tools/MintPy && \
+    git clone https://github.com/yunjunz/PyAPS.git /home/micromamba/tools/PyAPS
```
jhkennedy, line 23: I think you can switch back to the micromamba user before these and you'll avoid any potential ownership issues -- these are being owned by root, and then environment below is being created by root

jhkennedy: you've also hard-coded `micromamba` "user" here, but not in the ARGs below

forrestfwilliams: Thank you @jhkennedy , I see what you mean about the username, I'll go hard-code `micromamba` as the user and fix the other points you mentioned.

docker/Dockerfile
```diff
@@ -1,72 +1,47 @@
...
+# Set user name (the micromamba image comes with the micromamba user)
+ARG user=micromamba
 
-# Change default shell from dash to bash permanently
-SHELL ["/bin/bash", "-c"]
+# Clone MintPy and PyAPS repositories
+USER root
+RUN apt-get update && \
+    apt-get install -y git vim && \
+    rm /var/lib/apt/lists/* && \
```
jhkennedy, line 20:
```suggestion
    rm -rf /var/lib/apt/lists/* && \
```

jhkennedy: or at least `-r`

jhkennedy (commented):
This looks pretty good! And should be a big build improvement

forrestfwilliams:
Up to date version of the image can be found at forrestwilliams/mintpy:1.3.1

jhkennedy:
@forrestfwilliams this is great! 

I was able to build this image and run the tests inside it just fine, like:
```

docker run -it forrestwilliams/mintpy:1.3.1
cd tools/MintPy/tests
micromamba install -c conda-forge wget  # needed for test
python test_smallbaselineApp.py
```

So in terms of the dockerfile, I'd call this g2g! (unless it's worth adding `wget` to the `apt-get install` as well)

@yunjunz, I assume you'd like to do the final approval?

jhkennedy:
Other thoughts for you @forrestfwilliams  and @yunjunz:
*  in [`docker.md `](docs/docker.md) and [`installation.md`](dcos/installation.md), the ` andretheronsa/mintpy:latest` image is referenced, but it hasn't been updated in two years: 
  https://hub.docker.com/r/andretheronsa/mintpy
  I'd say it's worth updating to @forrestfwilliams image
 
* In a *subsequent* PR, it might be worth setting up a GitHub Action which will publish docker images to the GittHub container registry (as a package associated with this repo) for any push to main, so anyone could pull an up to date image like:
  ```
  docker pull ghcr.io/insarlab/mintpy:latest
  ```
  And probably also push a release tagged image (e.g., 1.3.1) for any release. If there's interest, I could team up with you @forrestfwilliams to set that up.

yunjunz:
Good point @jhkennedy. Let's update the docker image's link in the two *.md files to `forrestwilliams/mintpy`, could you please do that @forrestfwilliams?

docker/Dockerfile
```diff
@@ -1,72 +1,45 @@
...
-    apt-get install -y \
-    git \
-    python3 \
-    python3-pip \
-    wget \ 
-    vim \
-    zip \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    apt-get install -y git vim && \
```
yunjunz, line 16: I would suggest adding `wget` here since it is needed for running the example datasets, as @jhkennedy suggested.

docker/Dockerfile
```diff
@@ -1,72 +1,45 @@
...
-    ./Miniconda3-4.5.4-Linux-x86_64.sh -b -p ${PYTHON3DIR} && \
-    rm ./Miniconda3-4.5.4-Linux-x86_64.sh
-
-# Install dependencies
-RUN ${PYTHON3DIR}/bin/conda config --add channels conda-forge && \
-    ${PYTHON3DIR}/bin/conda install --yes --file ${MINTPY_HOME}/docs/requirements.txt
-
-# Install pykml
-RUN ${PYTHON3DIR}/bin/pip install git+https://github.com/yunjunz/pykml.git
+ENV PYTHONPATH=${MINTPY_HOME}:${PYAPS_HOME}
```
yunjunz, line 35: Since `MINTPY_HOME` and `PYAPS_HOME` are only used in the building and not needed afterward. We could use the `ARG` only, without `ENV` for them.

docker/Dockerfile
```diff
@@ -1,72 +1,45 @@
...
+WORKDIR /home/micromamba
 
-# Prepare build time environment for conda and mintpy
-ARG MINTPY_HOME=/home/python/MintPy
-ARG PYTHON3DIR=/home/python/miniconda3
-ARG PYAPS_HOME=/home/python/PyAPS
-ARG WORK_DIR=/home/work/
+RUN mkdir /home/micromamba/tools && \
+    git clone https://github.com/insarlab/MintPy.git /home/micromamba/tools/MintPy && \
+    git clone https://github.com/yunjunz/PyAPS.git /home/micromamba/tools/PyAPS
```
yunjunz, line 25: A not important detail: maybe define ARG MINTPY_HOME and PYAPS_HOME before the git clone and use them here to make it shorter?

yunjunz:
I tried to test the docker image following the same command as @jhkennedy wrote above. It passed the first two datasets for (ISCE and ARIA) and failed at generating KMZ file for the GAMMA dataset, as shown in the error message below. Manually running the `save_kmz.py` command works fine actually. Any idea why?

```bash
******************** step - google_earth ********************
creating Google Earth KMZ file for geocoded velocity file: ...

save_kmz.py /home/micromamba/data/test/WellsEnvD2T399/mintpy/geo/geo_velocity.h5 -o /home/micromamba/data/test/WellsEnvD2T399/mintpy/geo/geo_velocity.kmz
data   coverage in y/x: (0, 0, 753, 537)
subset coverage in y/x: (0, 0, 753, 537)
update LENGTH, WIDTH, Y/XMAX
update/add SUBSET_XMIN/YMIN/XMAX/YMAX: 90/12/843/549
update Y/X_FIRST
update REF_Y/X
read mask from file: geo_maskTempCoh.h5
masking out pixels with zero value in file: None
colormap: jet
plotting data ...
figure size : [16.83, 12.00]
show reference point
writing /home/micromamba/data/test/WellsEnvD2T399/mintpy/geo/geo_velocity.png with dpi=600
Killed
  File "test_smallbaselineApp.py", line 209, in <module>
    main(sys.argv[1:])
  File "test_smallbaselineApp.py", line 190, in main
    test_isce=inps.test_isce)
  File "test_smallbaselineApp.py", line 153, in test_dataset
    raise RuntimeError('Test failed for example dataset {}'.format(dset_name))
RuntimeError: Test failed for example dataset WellsEnvD2T399
```

jhkennedy:
@yunjunz the `Killed` in the line makes me suspect a memory issue (hitting a docker imposed limit?) -- I was able to run all of them through on my linux box.

yunjunz:
@jhkennedy you are right. After increasing the resource limit for Docker from the default values, all tests passed for me as well.

forrestfwilliams:
Morning! I'll go ahead and make these changes:

- add `wget`
- update `installation.md` and `docker.md`

@yunjunz, for `PYAPS_HOME` and `MINTPY_HOME` are you saying the user doesn't need to access them once MintPy is built? If so, I might go ahead and hard-code them.

yunjunz:
@forrestfwilliams yes, users do not need to access the two variables.

forrestfwilliams:
I've updated the dockerfile/docs. I also ran the tests in a container built from the new image and all tests passed! I'm re-uploading the image to docker and I'll let you know when it's done.

forrestfwilliams:
New build is available at `forrestwilliams/mintpy:1.3.1` !

forrestfwilliams:
Oops sorry I missed these

yunjunz:
No worries at all.

I am not a regular user of docker. @forrestfwilliams could you confirm if my modification of using ARG MINTPY_HOME and PYAPS_HOME in the DockerFile is reasonable?

yunjunz (approved):
Looks all good to me. 

Thank you @forrestfwilliams for updating this and @jhkennedy for the constructive review!

forrestfwilliams:
@yunjunz, I'd like to get @jhkennedy's opinion on this but my understanding is that `ARG`s should be reserved for variables that we want to be modifiable by someone building the image (see [here](https://docs.docker.com/engine/reference/builder/#arg)). Your modifications improve readability, but I'm not sure if we want the `MINTPY_HOME` and `PYAPS_HOME` locations to be modifiable. Thoughts?

docker/Dockerfile
```diff
@@ -1,72 +1,43 @@
...
-# Pull and config PyAPS
-RUN git clone https://github.com/yunjunz/PyAPS.git
-
-# Install miniconda
-RUN wget https://repo.continuum.io/miniconda/Miniconda3-4.5.4-Linux-x86_64.sh && \
-    chmod +x ./Miniconda3-4.5.4-Linux-x86_64.sh  && \
-    ./Miniconda3-4.5.4-Linux-x86_64.sh -b -p ${PYTHON3DIR} && \
-    rm ./Miniconda3-4.5.4-Linux-x86_64.sh
+# Download source code
+RUN mkdir /home/micromamba/tools && \
```
jhkennedy, line 30:
```suggestion
RUN mkdir -p ${MINTPY_HOME} ${PYAPS_HOME} && \
```

docker/Dockerfile
```diff
@@ -1,72 +1,43 @@
...
-    git \
-    python3 \
-    python3-pip \
-    wget \ 
-    vim \
-    zip \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    apt-get install -y git vim wget && \
+    rm -rf /var/lib/apt/lists/*
```
jhkennedy, line 17:
```suggestion
    apt-get install -y --no-install-recommends git vim wget && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```

Not necessary, but would help reduce the size of the image and speed up builds.

jhkennedy (commented):
@forrestfwilliams @yunjunz `ARG` is something that could be overridden at build time, but is also a convenient and often used way to have constants/variables in the docker file, so I don't see any issues using it for that. 

I only see one *small* problem is someone *did* specify different values at build time (the directory to clone into might not exist), but added a suggestion that would solve that.

docker/Dockerfile
jhkennedy: If someone did change the `ARG` for `MINTPY_HOME` or `PYAPS_HOME`, the parent directory might not exist for the subsequent clone commands. this insures the directories all exist.

yunjunz: Thank you @jhkennedy, this looks good to me.

yunjunz:
The PR is ready to merge in my opinion. @forrestfwilliams please feel free to merge it if you think it's ready.

forrestfwilliams:
Thanks @yunjunz ! I think we're ready to merge as well but I'm not sure I have the permissions to do so.

yunjunz:
I just sent you an invite. Could you try again after acceptance?

forrestfwilliams:
Merged! Thanks for inviting me to join insarlab!

yunjunz:
Great to have you here!

PR #687, opened 2021-11-03
spatial_filter: refactor double difference

forrestfwilliams:
- combine the local and regional filter into
  one filter to reduce number of
  convolutional operations

- change regional filter shape from a disk to
  a doughnut

**Reminders**

- [x] Fix #xxxx

### DIFF
--- a/mintpy/spatial_filter.py
+++ b/mintpy/spatial_filter.py
@@ -116,13 +116,16 @@ def filter_data(data, filter_type, filter_par=None):
     elif filter_type == "double_difference":
         """Amplifies the local deformation signal by reducing the influence
         of regional deformation trends from atmospheric artifacts, tectonic
-        deformation, and other sources. Intend use is to identify landslide-related
+        deformation, and other sources. Intended use is to identify landslide-related
         deformation. Filter has the form:
 
             result =  regional mean of data - local mean of data
 
         where both the regional and local kernel size can be set using the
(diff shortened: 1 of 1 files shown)


yunjunz:
Thank you @forrestfwilliams. Applying one convolution sounds like a good idea. I am wondering what's the motivation for changing the shape of the regional filter from a disk to a donut?

forrestfwilliams:
Hey @yunjunz, good question! The small `local_kernel` is meant to represent the local deformation signal, while the `regional_kernel` is supposed to represent the deformation occurring outside of the local area. By not including `local_kernel` pixels within the `regional_kernel` (i.e. making the regional kernel a doughnut) we are trying to capture only the regional signal in the `regional_kernel`.

In practice, @dbekaert found that the `regional_kernel` is typically large enough that the effect of including the `local_kernel` pixels is negligible, but I thought it would be good to make this modification regardless. Especially since it does not require any extra steps when you're already combining the two kernels.

yunjunz:
Thank you for the explanation @forrestfwilliams, that makes sense. Could you please add this as comments next to the code as well? I believe that would help other users too.

forrestfwilliams:
Can do!

yunjunz (approved):
Looks great. Thank you @forrestfwilliams.

forrestfwilliams:
Thank you @yunjunz !

PR #690, opened 2021-11-07
prep_hyp3: ignore whitespace in keys

forrestfwilliams:
In the metadata file provided with HyP3 InSAR products, the syntax of the UTC time key has changed form `UTCtime` to `UTC time` and the prep_hyp3 scripts needs to be updated to reflect this.

A try except statement was added that first tries to index on `UTC time` , then indexes on `UTCtime` if this produces a `Key Error`. This ensures backward compatibility with old products.

### DIFF
--- a/mintpy/prep_hyp3.py
+++ b/mintpy/prep_hyp3.py
@@ -117,7 +117,12 @@ def add_hyp3_metadata(fname,meta,is_ifg=True):
 
     # add universal hyp3 metadata
     meta['PROCESSOR'] = 'hyp3'
-    meta['CENTER_LINE_UTC'] = hyp3_meta['UTCtime']
+
+    try:
+        meta['CENTER_LINE_UTC'] = hyp3_meta['UTC time']
+    except KeyError:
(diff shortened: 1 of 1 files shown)


yunjunz:
Thank you for the update @forrestfwilliams. The subtle change of adding/removing whitespace in the upstream (HyP3) can be error-prone to the downstream code (MintPy). How about we ignore the whitespace in all keys while reading the HyP3 metadata file?

If so, a few other places in prep_hyp3.py need to remove the whitespace as well.

forrestfwilliams:
Yeah @yunjunz I think that would be an improvement. I'll go ahead and make the changes.

jhkennedy:
Sorry -- this one is my bad (I reviewed the HyP3 PR that made the change). Eventually we'd like to do something *a bit* better with an included metadata file, but for now, stripping whitespace is probably a good solution. We don't have plans in the near future to modify this file, or the keys in it however.

yunjunz (approved):
Thank you @forrestfwilliams for the update. It looks all good to me.

PR #693, opened 2021-11-10
plot_network: color-code network/matrix plot based on temp/perp baselines

yuankailiu:

Here are two proposed changes:

1). The displacement map in `tsview.py` needs to be projected onto a specific Cartopy projection (e.g., ccrs.PlateCarree()) to properly overlay the coastline. This is added. But I am not sure whether there should be tunable projections. For small regions (for most of the cases), I think it is fine.

2). In 'plot_network.py', Other than coherence, I personally think it would be good to also allow the lines (connecting acquisitions) to be colored based on their temporal or spatial baselines. This is just for network design displaying purposes. You can reject that as well.

**Reminders**

### DIFF
--- a/mintpy/plot_network.py
+++ b/mintpy/plot_network.py
@@ -37,6 +37,7 @@ EXAMPLE = """example:
   plot_network.py inputs/ifgramStack.h5
   plot_network.py inputs/ifgramStack.h5 -t smallbaselineApp.cfg --nodisplay   #Save figures to files without display
   plot_network.py inputs/ifgramStack.h5 -t smallbaselineApp.cfg --show-kept   #Do not plot dropped ifgrams
+  plot_network.py inputs/ifgramStack.h5  --tb --tbv 0 365.25 -c2 RdYlBu_r  #Color lines by temporal baseline with -c2 cmap
   plot_network.py coherenceSpatialAvg.txt
 
   # offsetSNR
@@ -75,6 +76,19 @@ def create_parser():
 
     coh.add_argument('-v', '--vlim', nargs=2, type=float, default=(0.2, 1.0),
                      help='display range')
+
+    # Display baseline in the network plot
+    coh = parser.add_argument_group('Display Baseline', 'Color lines in the network plot with baseline instead of coherence')
+    coh.add_argument('--tb', '--tbase', dest='tbColor', action='store_true',
+                     help='color using temporal baseline')
+    coh.add_argument('--pb', '--pbase', dest='pbColor', action='store_true',
+                     help='color using perpendicular baseline')
+    coh.add_argument('--tbv', '--tbvlim', dest='tbvlim', nargs=2, type=float, default=(0, 365.25),
+                     help='display color range for temporal baseline [day]. (default: %(default)s)')
+    coh.add_argument('--pbv', '--pbvlim', dest='pbvlim', nargs=2, type=float, default=(0, 180.0),
+                     help='display color range for perpendicular baseline [meter]. (default: %(default)s)')
+    coh.add_argument('-c2', '--colormap2', dest='cmap_name2', default='RdYlBu_r',
+                     help='colormap name for the network display (baseline coloring). Default: %(default)s')
 
     # Figure  Setting
     fig = parser.add_argument_group('Figure', 'Figure settings for display')
@@ -233,6 +247,7 @@ def check_colormap(inps):
     # in case the manually input list is not in order
     inps.cmap_vlist = sorted(inps.cmap_vlist)
     inps.colormap = pp.ColormapExt(inps.cmap_name, vlist=inps.cmap_vlist).colormap
+    inps.colormap2 = pp.ColormapExt(inps.cmap_name2, vlist=inps.cmap_vlist).colormap
 
     return inps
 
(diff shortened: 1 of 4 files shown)


yunjunz (changes requested):
Thank you @yuankailiu, the two features sound fine to me.

Here are my suggestions:

1. For the `plot_network.py`, I tried to run `plot_network.py inputs/ifgramStack.h5 --tb` and it does seem to work. Could you double-check, and add an example usage as well?

2. For the `tsview.py`, it's more complicated, as the script needs to support files in radar coordinates as well. Below is how `view.py` handles this, which eventually translated into the `inps.map_proj_obj` variable. I would recommend: packing this checking into a sub-function in view.py or utils/plot.py, and call it from tsview.py. If you want to go for this, please consider submitting it in another PR, as it requires much more changes.

https://github.com/insarlab/MintPy/blob/92e00ad521e05adb1bff3130cd97b1d00e9cb8d7/mintpy/view.py#L310-L345

yuankailiu:
Hi Yunjun, thanks for reviewing this.

`plot_network.py` and `plot.py` have been updated, with an example added in `plot_network.py -h`. Now it is working after my testing.

I have also changed the `utils1.py` a bit in the `get_residual_std()` function. I make it to return the `std_file` as the `get_residual_rms()` function.

As for `tsview.py`, I will do it later in another PR. Thanks!

mintpy/utils/utils1.py
```diff
@@ -86,7 +86,7 @@ def get_residual_std(timeseries_resid_file, mask_file='maskTempCoh.h5', ramp_typ
     fc = np.loadtxt(std_file, dtype=bytes).astype(str)
     std_list = fc[:, 1].astype(np.float32).tolist()
     date_list = list(fc[:, 0])
-    return std_list, date_list
+    return std_list, date_list, std_file
```
yunjunz, line 89: Sounds good to me. Could you also update the related comments in the following lines?

https://github.com/insarlab/MintPy/blob/4357b8c726dec8a3f936770e3f3dda92882685b7/mintpy/utils/utils1.py#L54-L58

mintpy/timeseries_rms.py
```diff
@@ -222,8 +222,8 @@ def main(iargs=None):
     (inps.rms_list,
      inps.date_list,
      inps.rms_file) = ut.get_residual_rms(inps.timeseries_file,
-                                          mask_file=inps.maskFile,
-                                          ramp_type=inps.deramp)
+                                         mask_file=inps.maskFile,
+                                         ramp_type=inps.deramp)
```
yunjunz, line 226: The changes in this file seem unnecessary, could we undo them?

mintpy/plot_network.py
```diff
@@ -76,6 +77,19 @@ def create_parser():
     coh.add_argument('-v', '--vlim', nargs=2, type=float, default=(0.2, 1.0),
                      help='display range')
 
+    # Display baseline in the network plot
+    coh = parser.add_argument_group('Display Baseline', 'Color lines in the network plot with baseline instead of coherence')
```
yunjunz, line 81: `coh` name is taken for the previous argument group, let's rename it to something else, e.g. `base`?

mintpy/plot_network.py
```diff
@@ -76,6 +77,19 @@ def create_parser():
...
+    coh.add_argument('--tb', '--tbase', dest='tbColor', action='store_true',
+                     help='color using temporal baseline')
+    coh.add_argument('--pb', '--pbase', dest='pbColor', action='store_true',
+                     help='color using perpendicular baseline')
+    coh.add_argument('--tbv', '--tbvlim', dest='tbvlim', nargs=2, type=float, default=(0, 365.25),
+                     help='display color range for temporal baseline [day]. (default: %(default)s)')
+    coh.add_argument('--pbv', '--pbvlim', dest='pbvlim', nargs=2, type=float, default=(0, 180.0),
+                     help='display color range for perpendicular baseline [meter]. (default: %(default)s)')
+    coh.add_argument('-c2', '--colormap2', dest='cmap_name2', default='RdYlBu_r',
```
yunjunz, line 90: It seems like a new `--colormap2` option is not necessary, we could use/share the same `-c` for this purpose. Do you agree?

yunjunz (changes requested):
Thank you @yuankailiu for the updates. Could you check the few minor suggestions I listed above?

After that, the code is clean enough and I could work on the idea below (unless you are willing to work that, in that case let me know), which requires quite some refactoring:

Your implementation makes me think that we could use a more generic/flexible `--dset dsName` option: if `dsName` is a 3D matrix, then use `maskFIle` to calculate the spatial average (which is what currently do); if `dsName` is a 1D matrix, e.g. bperp, treat it as the calculated spatial average directly, and all the following code should be the same.

mintpy/plot_network.py
yuankailiu: Ok, this and the following related lines have been removed.

mintpy/timeseries_rms.py
yuankailiu: Sure. I am sorry. This is my VSCode automatically identifies and removes the spaces that are not consistent with the other indentations.

mintpy/utils/utils1.py
yuankailiu: The comment is added there. Thanks.

yunjunz (approved):
Looks all good to me.

PR #699, opened 2021-11-18
coord.lalo2yx(): replace np.float128 with np.longdouble for Windows

### LINKED ISSUE #698: AttributeError: module 'numpy' has no attribute 'float128' in coord.py
I just updated my MintPy and now I have this error where NumPy has no attribute `float128`.
It comes from this line:
https://github.com/insarlab/MintPy/blob/d2727e06b8fea2d74ff870e5c989bbd9485515eb/mintpy/objects/coord.py#L89

I looked into it and it seems that float128 is not supported on Windows OS. I removed `np.float128` from the line and my analysis works with no error.

See related Github issue from other repo for reference: https://github.com/winpython/winpython/issues/613

If precision is still required up to float128 then here is a possible solution: https://github.com/quantumlib/Cirq/issues/1511#issuecomment-481450449

pbrotoisworo:

Simple change that modifies `np.float128` object to `np.longdouble`. This is because float128 is not supported on Windows OS and will raise errors for Windows users.

**Reminders**

- [ ] Fix #698

### DIFF
--- a/mintpy/objects/coord.py
+++ b/mintpy/objects/coord.py
@@ -86,6 +86,7 @@ class coordinate:
         # input format
         if isinstance(coord_in, np.ndarray):
             coord_in = coord_in.tolist()
+        # note: np.float128 is not supported on Windows OS, use np.longdouble as a platform neutral syntax
         if isinstance(coord_in, (float, np.float16, np.float32, np.float64, np.longdouble)):
             coord_in = [coord_in]
         coord_in = list(coord_in)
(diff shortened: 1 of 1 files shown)


pbrotoisworo:
Hmm something is wrong on CircleCI during the installation process (timed out)? I don't think it's related to the code I changed.

yunjunz (approved):
LGTM. Thank you @pbrotoisworo for the fix. 

You are right: there is an issue with circle CI at the moment, it's not related to your change.

PR #700, opened 2021-11-19
update installation note, env setup and circle-ci testing

jhkennedy:
Since #659 , there's no longer a need to manipulate `PATH` or `PYTHONPATH` as a `pip` install will correctly place the `mintpy` package and all executable scripts to be available in the environment. 

Additionally, PyPAS and PySolid are available via conda-forge and no longer need to be cloned and manually built/installed. 

*Draft because I need to do docs yet*

**Reminders**

- [x] Fix #xxxx

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -28,7 +28,6 @@ dependencies:
   - pyaps3
   # for ARIA, FRInGE, HyP3, GMTSAR
   - gdal>=3
-  # - pyhdf   # required by MERRA, which is currently not supported in pyaps3
   # for pyresample
   - pyresample
   - openmp
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -20,12 +20,12 @@ dependencies:
   - lxml
   - matplotlib
   - numpy
+  - pyaps3
   - pykml>=0.2
   - pyproj
+  - pysolid
   - scikit-image
   - scipy
-  - pysolid
-  - pyaps3
   # for ARIA, FRInGE, HyP3, GMTSAR
   - gdal>=3
   # for pyresample
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -9,6 +9,7 @@ channels:
   - defaults
 dependencies:
   - python>=3.6,<3.9
+  - pip
   - cartopy
   - cvxopt
   - dask>=1.0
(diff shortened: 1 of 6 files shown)


docs/environment.yml
```diff
@@ -19,20 +20,14 @@ dependencies:
...
+  - pysolid
+  - pyaps3
   # for ARIA, FRInGE, HyP3, GMTSAR
   - gdal>=3
-  # for PyAPS
-  - cdsapi
-  - ecCodes
-  - netcdf4
-  - pygrib
   # - pyhdf   # required by MERRA, which is currently not supported in pyaps3
```
yunjunz, line 31:
```suggestion
```

docs/requirements.txt
```diff
@@ -14,21 +15,18 @@ joblib
...
-ecCodes
-netcdf4
-pygrib
-# pyhdf  # required by MERRA, which is currently not supported in pyaps3
-# pyresample dependencies
+pysolid
+pyaps3
+# for ARIA, FRInGE, HyP3, GMTSAR
+gdal>=3
+# - pyhdf   # required by MERRA, which is currently not supported in pyaps3
```
yunjunz, line 26:
```suggestion
```

Let's remove this line for cleanness.

docs/requirements.txt
```diff
@@ -14,21 +15,18 @@ joblib
...
-cdsapi
-ecCodes
-netcdf4
-pygrib
-# pyhdf  # required by MERRA, which is currently not supported in pyaps3
-# pyresample dependencies
+pysolid
+pyaps3
+# for ARIA, FRInGE, HyP3, GMTSAR
+gdal>=3
```
yunjunz, line 25: `gdal` is still not so easy to install and is not required if one is not using the above InSAR products. Since `requirements.txt` is meant for installing mintpy to an existing env, I would prefer keeping gdal out of the `requirements.txt` file, for more flexibility.

It's easy to add gdal in the command line while running conda install if one wants to.

yunjunz:
It's great that the conda install `pysolid` and `pyaps3` is fully working! Thank you @jhkennedy!

I could help with the doc updates, will try it this weekend.

docs/requirements.txt
jhkennedy: :+1: makes sense to me -- I'll leave it here, but commented out.

jhkennedy:
@yunjunz I took a first stab at the documentation updates. `conda` install should be the same on any platform now, so I removed the windows install note. 

Also, I've update the `pip` commands to use the preferred `python -m pip` invocation (long discussion at pypa/pip#3164), except for macports because I'm don't use a mac. I suspect for macports you'd also just do a `sudo -H python3 -m pip`. 

I'm off to a cabin for the weekend tomorrow, so feel free to run with this, or I'll be back to iterate on Monday. :smile:

docs/requirements.txt
jhkennedy: Well, I commented it out, but because CircleCI uses `requirements.txt` and the `SanFranSenDT42` test needs GDAL, I had to add it to the CicleCI config. 

While `gdal` can be a rather large pain, in general installing it via conda is pretty painless at this point, so I don't see a big harm including it. 

I'm not terribly invested either way though as I will almost always use the `environment.yml` -- whichever way you want to go is fine by me.

yunjunz (approved):
Looks all good to me now!

For the macports part, since I don't use macports actively anymore, I did not change the previous `pip install` style.

PR #704, opened 2021-11-23
set matplotlib<3.5 temporarily due to upstream repo bug

yunjunz:

set matplotlib<3.5 temporarily due to upstream repo bug (https://github.com/matplotlib/matplotlib/issues/21711), which leads error while saving figures to PDF files.

**Reminders**

- [x] Pass Circle CI / local test (green)

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -18,7 +18,7 @@ dependencies:
   - h5py<3
   - joblib
   - lxml
-  - matplotlib
+  - matplotlib<3.5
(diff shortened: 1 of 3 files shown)


jhkennedy (approved):
Looks good to me! Might be worth making an issue for unpinning this dependency so we don't for get about it in the future.

yunjunz:
Good idea @jhkennedy, I opened a new issue for that as reminders.

PR #719, opened 2022-01-11
add `mintpy.unwrapError.connCompMinArea` option

mirzaees:

Some of the parameters which were hardcoded, are change to be optional.

1. The step size in solid earth tide function (made adaptive based on input file info)
2. The min_area in label_conn_comp (added an option in template)
3. Added support for slcStack to read slice list

**Reminders**


yunjunz:
Thank you @mirzaees for the improvements.

Your change in the `readfile.get_slice_list()` won't affect existing functionality, which is good. I am just curious on your logic behind: we identify the `FILE_TYPE` via the h5 dataset name `timeseries` in the root level, and your change seems to assume a h5 dataset name `slc` in the root level, thus, I don't think it will work, isn't it?

Related to the above, I am curious how you distinguish between the displacement time series and the SLC stack in general in minopy?

mirzaees:
For slcStack, the dataset name is slc but the file type is timeseries, minopy has a slcStack.py module similar to mintpy stack.py to make it. File type was previously SLC too but some of the mintpy functions did not work so that is now changed to timeseries. One way to distinguish is data type as you mentioned and the dataset name is also different.

yunjunz (approved):
Thank you for the explanation @mirzaees. Looks all good to me.

PR #726, opened 2022-01-20
fix typo in dask documentation

olliestephenson:

`numWorkers` should read `numWorker` in dask documentation


yunjunz (approved):
Good catch!

PR #732, opened 2022-01-27
generic multi-dataset file (e.g. velocity) support in `add/diff/deramp`

yuankailiu:
+ add.py: when operating on velocity files, operate on velocity dset as default
+ diff.py: when operating on velocity files, operate on velocity dset as default
+ utils1..py: when deramping on velocity files, operate on velocity dset as default

Here propose to operate on the velocity dataset as default when using the following functions/codes on the velocity files (`FILE_TYPE`=velocity)

Before, the default for deramping is operating on the first dataset, not necessarily the "velocity" if there are multiple fields in it, e.g., "annualAmplitude".

Also the same situation in `add.py` and `diff.py`.

This is a quick tweak, it would be even better to allow a user-defined `dataset` when calling the function, which could be in another PR.

Thanks for reviewing it.

**Reminders**

### DIFF
--- a/mintpy/utils/utils1.py
+++ b/mintpy/utils/utils1.py
@@ -837,7 +837,9 @@ def run_deramp(fname, ramp_type, mask_file=None, out_file=None, datasetName=None
             with open(coeff_file, 'a') as f:
                 f.write('{}    '.format(atr['FILE_TYPE']))
         # read
-        data = readfile.read(fname)[0]
+        if not datasetName and k == 'velocity':
+            datasetName = 'velocity'
+        data = readfile.read(fname, datasetName=datasetName)[0]
         # deramp
         data = deramp(data, mask,
                       ramp_type=ramp_type,
--- a/mintpy/add.py
+++ b/mintpy/add.py
@@ -153,7 +153,7 @@ def add_file(fnames, out_file=None, force=False):
                 # ignore ds_name if input file has single dataset
                 ds_name2read = None if len(ds_names_list[i+1]) == 1 else ds_name
                 # read
-                data2 = readfile.read(fnames[i], datasetName=ds_name2read)[0]
+                data2 = readfile.read(fname, datasetName=ds_name2read)[0]
                 # apply operation
                 data = add_matrix(data, data2)
             dsDict[ds_name] = data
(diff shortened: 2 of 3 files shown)


mintpy/utils/utils1.py
```diff
@@ -830,6 +830,24 @@ def run_deramp(fname, ramp_type, mask_file=None, out_file=None, datasetName=None
...
+        data = deramp(data, mask,
+                    ramp_type=ramp_type,
+                    metadata=atr,
+                    coeff_file=coeff_file)[0]
+        # write
+        print('writing >>> {}'.format(out_file))
+        writefile.write(data, out_file=out_file, ref_file=fname)
+
     # Single Dataset File
     else:
```
yunjunz, line 852: Instead of creating a new `elif` block, we could use the existing `else` block by assigning the `datasetName` for the velocity file as below:

```python
data = readfile.read(fname)[0]
```

to

```python
if not datasetName and k == 'velocity':
    datasetName = 'velocity'
data = readfile.read(fname, datasetName=datasetName)[0]
```

yunjunz (changes requested):
Thank you @yuankailiu for the useful PR. Could we use the existing `else` block by assigning `dsName` or `datasetName` for the velocity file type, for this functionality, instead of creating a new `elif` block of code? That would be a more adaptable solution.

mintpy/utils/utils1.py
yuankailiu: Hi @yunjunz, thanks for the advice. I will do that!

yunjunz:
Hi @yuankailiu, we should apply the same style of `utils1.py` to the other two scripts as well.

yuankailiu:
> Hi @yuankailiu, thank you for addressing the suggestion. We should apply the same style of changes in `utils1.py` to the other two scripts as well.

@yunjunz, for `add.py` and `diff.py`, thanks. I have this new commit with a maybe more general way. It is still using the current `else` block. Now I first identify the common datasets for two (or more) input files and operate on those common datasets. 

So now `diff.py` imitates `add.py` to handle multiple datasets.

Do you think this is a good way to go?

yunjunz:
Hi @yuankailiu, that is a good idea.

yunjunz (approved):
Looks all good to me now.

PR #733, opened 2022-01-28
bugfix for `timeseries2velocity.py --bootstrap --save-res`

olliestephenson:
When running `timeseries2velocity.py` with bootstrapping, we use the function `time_func.estimate_time_func` to repeatedly fit the time series data. This function returns the design matrix, `G`,  however when bootstrapping we don't assign the design matrix to a variable, meaning that if we try to save the residual file we get the error `UnboundLocalError: local variable 'G' referenced before assignment` at line 702. 

Related to https://github.com/insarlab/MintPy/pull/654

If we want to output the residual when bootstrapping, we need to get the full design matrix. I've added a line of code to do this.

**Reminders**


yunjunz (approved):
LGTM. Thank you @olliestephenson.

Issue #705, opened 2021-11-23
temporary constraints on the dependencies max version number


yunjunz:
Add a reminder here to unset the following maximum version number constraints, due to temporary issues/bugs in the upstream dependencies.

+ [x] (un)set `h5py<3` due to https://github.com/h5py/h5py/issues/1750
   - #470 
   - #474 
   - https://github.com/insarlab/MintPy/pull/743
+ [x] (un)set `matplotlib<3.5`  due to https://github.com/matplotlib/matplotlib/issues/21711
    - #704
    - #706

PR #735, opened 2022-02-01
prep_fringe: write ifgramStack for connected components

bbuzz31:

Here a few updates for `prep_fringe`.

- you can specify the input argument ` --geom-only ` to only write out geometry files (useful for making masks)
- you can specify looks (this is only for tracking metadata)
- if it can find connected component files, it will make `ifgramStack.h5`.
  - it will find them as long as they're the same names as the unwrapped interograms (required input) except with `.conncomp` extension

**Reminders**

### DIFF
--- a/mintpy/prep_fringe.py
+++ b/mintpy/prep_fringe.py
@@ -135,10 +135,9 @@ def read_vrt_info(vrt_file):
     # source dir
     type_tag = root.find(prefix + '/SourceFilename')
     src_dir = os.path.dirname(type_tag.text)
-    if src_dir is None:
+    if not src_dir:
         src_dir = os.path.dirname(vrt_file)
     
-
     return src_box, src_dir
 
 
--- a/mintpy/prep_fringe.py
+++ b/mintpy/prep_fringe.py
@@ -368,6 +368,7 @@ def prepare_stack(outfile, unw_file, metadata, processor, baseline_dir=None, box
         baseline_dict = isce_utils.read_baseline_timeseries(baseline_dir, processor=processor)
 
         # calc baseline for each pair
+        print('calc perp baseline pairs from time-series')
         pbase = np.zeros(num_pair, dtype=np.float32)
         for i, date12 in enumerate(date12_list):
             [date1, date2] = date12.split('_')
@@ -500,7 +501,8 @@ def main(iargs=None):
         baseline_dir=inps.baselineDir,
         box=pix_box)
 
-    return ts_file, tcoh_file, ps_mask_file, geom_file
+    print('Done.')
+    return
 
 
 ####################################################################################
--- a/mintpy/prep_fringe.py
+++ b/mintpy/prep_fringe.py
@@ -76,9 +76,11 @@ def create_parser():
                         help='output directory (default: %(default)s).')
 
     parser.add_argument('-r','--range', dest='lks_x', type=int, default=1,
-                        help='number of multilooking in range  /x direction (default: %(default)s).')
+                        help='number of multilooking in range direction. '
+                             'Only impacts metadata. (default: %(default)s).')
     parser.add_argument('-a','--azimuth', dest='lks_y', type=int, default=1,
-                        help='number of multilooking in azimuth/y direction (default: %(default)s).')
+                        help='number of multilooking in azimuth direction. '
+                             'Only impacts metadata. (default: %(default)s).')
 
     parser.add_argument('--geom-only', action='store_true',
                         help='Only create the geometry file (useful for geocoding a watermask).')
@@ -446,7 +448,7 @@ def main(iargs=None):
     src_box, geom_src_dir = read_vrt_info(os.path.join(inps.geomDir, 'lat.vrt'))
 
     # metadata
-    meta = prepare_metadata(inps.metaFile, geom_src_dir, box=src_box, inps.lks_x, inps.lks_y)
+    meta = prepare_metadata(inps.metaFile, geom_src_dir, src_box, inps.lks_x, inps.lks_y)
 
 
     # subset - read pix_box for fringe file
--- a/mintpy/prep_fringe.py
+++ b/mintpy/prep_fringe.py
@@ -364,9 +364,10 @@ def prepare_stack(outfile, unw_file, corr_file, metadata, processor, baseline_di
     num_pair = arr_dates12.shape[0]
     print('number of interferograms:', num_pair)
 
-    # get connected component file extension from last ifg
-    cc_file  = f'{os.path.splitext(unw_file)[0]}.conncomp'.replace(
-                                        f'{ref_date}_{sec_dates[-1]}', '*')
+    # get all connected component files using an ifg 
+    cc_file  = f'{os.path.splitext(unw_file)[0]}.conncomp'
+    cc_file  = os.path.join(os.path.dirname(cc_file), 
+                                    f'*{os.path.basename(cc_file)[17:]}')
 
     cc_files = sorted(glob.glob(cc_file))
     if not cc_files:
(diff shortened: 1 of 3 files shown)


yunjunz:
The issue of h5py indexing has been listed here (https://github.com/insarlab/MintPy/issues/705). The current version should work if you install `h5py<3`. 

I planned to wait for h5py fixing, but it does not seem to happen anytime soon (and we are already in h5py version 3.6 ...). If we are gonna use different logic for h5py indexing, as suggested here (https://github.com/h5py/h5py/issues/1750), I would prefer we do it in another PR, for all the occurrences.

mintpy/prep_fringe.py
```diff
@@ -329,6 +438,15 @@ def main(iargs=None):
...
 
+    # optionally apply user multilooking
+    meta['AZIMUTH_PIXEL_SIZE'] = str(float(meta['AZIMUTH_PIXEL_SIZE']) * inps.lks_y)
+    meta['RANGE_PIXEL_SIZE'] = str(float(meta['RANGE_PIXEL_SIZE']) * inps.lks_x)
+    meta['azimuthPixelSize'] = meta['AZIMUTH_PIXEL_SIZE']
+    meta['rangePixelSize']   = meta['RANGE_PIXEL_SIZE']
+
+    meta['ALOOKS'] = str(float(meta['ALOOKS']) * inps.lks_y)
+    meta['RLOOKS'] = str(float(meta['RLOOKS']) * inps.lks_x)
+
```
yunjunz, line 449: Could we move this inside `prepare_metadata()`?

mintpy/prep_fringe.py
```diff
@@ -329,6 +438,15 @@ def main(iargs=None):
     # metadata
     meta = prepare_metadata(inps.metaFile, geom_src_dir, box=src_box)
 
+    # optionally apply user multilooking
+    meta['AZIMUTH_PIXEL_SIZE'] = str(float(meta['AZIMUTH_PIXEL_SIZE']) * inps.lks_y)
+    meta['RANGE_PIXEL_SIZE'] = str(float(meta['RANGE_PIXEL_SIZE']) * inps.lks_x)
+    meta['azimuthPixelSize'] = meta['AZIMUTH_PIXEL_SIZE']
+    meta['rangePixelSize']   = meta['RANGE_PIXEL_SIZE']
```
yunjunz, line 445: So far, all the mintpy functions have been only maintaining consistency for metadata listed here (https://mintpy.readthedocs.io/en/latest/api/attributes/), I would recommend removing these two lines for consistency, even though they are reasonable.

mintpy/prep_fringe.py
```diff
@@ -75,6 +75,14 @@ def create_parser():
     parser.add_argument('-o', '--out-dir', dest='outDir', type=str, default='./mintpy',
                         help='output directory (default: %(default)s).')
 
+    parser.add_argument('-r','--range','-x', dest='lks_x', type=int, default=1,
+                        help='number of multilooking in range  /x direction (default: %(default)s).')
+    parser.add_argument('-a','--azimuth','-y', dest='lks_y', type=int, default=1,
+                        help='number of multilooking in azimuth/y direction (default: %(default)s).')
+
```
yunjunz, line 82: 1. I would recommend not using `-x` and `-y` here because they are more commonly used as to specify the subset range in other scripts.

2. I don't quite get the purpose of this. Could you elaborate?

mintpy/prep_fringe.py
```diff
@@ -127,7 +135,9 @@ def read_vrt_info(vrt_file):
     # source dir
     type_tag = root.find(prefix + '/SourceFilename')
     src_dir = os.path.dirname(type_tag.text)
-
+    if not src_dir:
+        src_dir = os.path.dirname(vrt_file)
+
```
yunjunz, line 140: It would be helpful to write a comment here on the reason.

mintpy/prep_fringe.py
bbuzz31: sure, was curious as to why there was a duplicate

mintpy/prep_fringe.py
bbuzz31: 1. Yes, can swith to -r and -a like in multilook.py 
2. Right now it's only for adding the metadata, which I'm finding useful for keeping track of which timeseries corresponds to which experiment. 
- I have a set of bash scripts that prepares the mulitlooked metadata for `prep_fringe`. Eventually I hope to incorporate them so that when you specify looks to `prep_fringe` it will make them on the fly

mintpy/prep_fringe.py
bbuzz31: When I create a multilooked vrt the fields are slightly off and that XML tag returns empty. I can add a comment.

bbuzz31:
> 

K, I can remove the two commits that implement these.

mintpy/prep_fringe.py
yunjunz: Do you mean that you have some multilooked data files, based on fringe outputs, but then the metadata is not consistent with the data file, so you would like to update them here?

bbuzz31: Exactly

yunjunz: `multilook.py -x -y` looks clear and intuitive to me, as it only does the multilooking, and applies to geocoded files. The same also applies to `subset.py -x -y`. While `prep_fringe.py -x -y` can be ambiguous, as we don't know it's for multilooking or subset. I hope this explains that I am not double standard ;)

yunjunz: Thanks for the explanation. That makes sense. Please add a comment to the help msg as well.

bbuzz31: Thanks no prob!

mintpy/prep_fringe.py
```diff
@@ -318,6 +341,105 @@ def prepare_geometry(outfile, geom_dir, box, metadata):
...
+    # define (and fill out some) dataset structure
+    dropIfgram = np.ones(arr_dates12.shape[0], dtype=bool)
+    ds_name_dict = {
+        "date"       : [np.dtype('S8'), (num_pair, 2), arr_dates12],
+
+        "bperp"      : [np.float32,  (num_pair,), pbase],
+        "dropIfgram" : [np.bool_,    (num_pair,), dropIfgram],
+
+        "coherence"  : [np.float32,  (box[3]-box[1], box[2]-box[0]), corr],
```
yunjunz, line 410: `coherence` dataset is optional for `ifgramStack`. Since we have only one temporal coherence here, I would recommend not loading coherence at all.

mintpy/prep_fringe.py
```diff
@@ -318,6 +341,105 @@ def prepare_geometry(outfile, geom_dir, box, metadata):
...
+    print('number of associated connected components:', len(cc_files))
+
+    # baseline info
+    if baseline_dir is not None:
+        # read baseline data
+        baseline_dict = isce_utils.read_baseline_timeseries(baseline_dir,
+                                                            processor=processor,
+                                                            ref_date=ref_date)
+        pbase = np.array([baseline_dict[sec][0] for sec in sec_dates],
+                                                            dtype=np.float32)
```
yunjunz, line 386: `pbase` here assumes the 1st date as the reference date. This is true for the current version of fringe output, but is likely to change in the near future. Could you re-construct the pbase for each pair, based on the 1st date of date12, to be more agnostic?

bbuzz31: good suggestion; i believe the last commit takes care of this

yunjunz (approved):
Testing passed on my local example fringe dataset. 

Thank you @bbuzz31 for the useful feature!

bbuzz31:
Great! Probably not all that useful for anyone but me yet, but hopefully with a few more future PRs it will be :D

PR #737, opened 2022-02-04
resample: bugfix for auto_lat_lon_step_size() via merged src/lut_meta dict

### LINKED ISSUE #736: Unexpected changes while geocoding
I've run into two issues regarding commit 70d606ee

1. I can no longer geocode a watermask (created iwth `generate_mask.py` ) as it doesn't have the heading attribute:
-```/mintpy/utils/utils0.py", line 222, in auto_lat_lon_step_size
    az_angle = np.deg2rad(abs(heading2azimuth_angle(float(atr['HEADING']))))
KeyError: 'HEADING'```

2. The size of the geocoded files (for example temporal coherence mask) is no longer consistent with the original radar coded mask (and I think width/height are flipped). Prior to this commit, the sizes were consistent. I'm not sure if this is expected or not. 
-the original size was (1504 x 3177) and is now (3213 x 1798).

Any thoughts on fixes? I can help implement.

bbuzz31:

This fixes #736 by passing all possible metadata to the geocoding functions.

**Reminders**

### DIFF
--- a/mintpy/objects/resample.py
+++ b/mintpy/objects/resample.py
@@ -338,11 +338,10 @@ class resample:
 
             # parameter 1 - lalo_step (output grid)
             if self.lalo_step is None:
-                dct_meta = self.lut_meta.copy()
-                dct_meta.update(self.src_meta)
+                dct_meta           = {**self.lut_meta, **self.src_meta}
                 lat_c              = (src_lat0 + src_lat1) / 2.
                 lat_step, lon_step = ut.auto_lat_lon_step_size(dct_meta, lat_c)
-                self.lalo_step = (abs(lat_step) * -1., abs(lon_step))
+                self.lalo_step     = (abs(lat_step) * -1., abs(lon_step))
 
             else:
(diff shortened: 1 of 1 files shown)


mintpy/objects/resample.py
```diff
@@ -338,13 +338,10 @@ def mark_lat_lon_anomoly(lat, lon):
 
             # parameter 1 - lalo_step (output grid)
             if self.lalo_step is None:
-                try:
-                    lat_c = (src_lat0 + src_lat1) / 2.
-                    lat_step, lon_step = ut.auto_lat_lon_step_size(self.src_meta, lat_c)
-                except AttributeError:
```
yunjunz, line 344: We should keep the `try/except` logic, in case neither lut_meta nor src_meta has the metadata needed for `ut.auto_lat_lon_step_size()`, if `except KeyError:` still works as you described in #736.

yunjunz (approved):
Thank you for the fixing @bbuzz31!

PR #738, opened 2022-02-04
save_kite.py - convert to meters for UNIT in radians

mgovorcin:

**save_kite.py** - minor fixes for: 1) if .unw is used as input, 2) if input file UNIT is radian, convert to meters

**coherenceCD.py** - very simple script that generates a coherence change map as a difference between the average of coherence stack before the event (earthquake, vulcano eruption itd..) and average of coherence stack covering the event.
Below is the example for the Fernandina test dataset (vulcano eruption Sep 2017)

Screen Shot 2022-02-03 at 9 32 59 PM

Please see if you find it useful

**Reminders**

- [ ] Fix #xxxx


mintpy/save_kite.py
```diff
@@ -158,9 +158,12 @@ def main(iargs=None):
 
     elif attr['FILE_TYPE'] == 'ifgramStack':
         date1, date2 = inps.dset.split('-')[1].split('_')
-
+        
+    elif attr['FILE_TYPE'] == '.unw':
+        date1, date2 = ptime.yyyymmdd(attr['DATE12'].split('-'))
+
```
yunjunz, line 164: It seems to me that we don't need a separate `elif` check for `.unw` as it's covered in the `else` block below.

yunjunz:
Thank you @mgovorcin for maintaining `save_kite.py`.

Regarding `coherenceCD.py`, it's a great demonstration of using mintpy as a python module for customized work! I think it fits better as an example notebook on [`MintPy-tutorial`](https://github.com/insarlab/MintPy-tutorial#4-custom-applications) repo, rather than an executable script within mintpy, because this procedure is not a standard processing step, nor it is peer-reviewed in publications (correct me on this if I am wrong please), plus we are moving mintpy toward a direction with less executable scripts, and more modules/functions in python level as APIs.

I would suggest keeping this PR for `save_kite.py` fixing only, and converting the `coherenceCD.py` into a jupyter notebook for a PR in `mintpy-tutorial` if you like.

mintpy/save_kite.py
mgovorcin: @yunjunz you are right, these lines are redundant...I will delete them

mgovorcin:
Thank you @yunjunz for your comment on coherenceCD.py, I will gladly convert this script to an example jupyter notebok and submit it to MintPy-tutorial repo. Hope some people might find it useful

yunjunz (approved):
LGTM. Thank you @mgovorcin.

PR #740, opened 2022-02-07
closure_phase_bias.py - mask for areas susceptible to closure phase bias

yjzhenglamarmota:

Create a mask for areas susceptible to closure phase bias.

I added a standalone script - creatmask_cp.py - to enable creation of a mask that identifies areas susceptible to closure phase bias.


mintpy/createmask_cp.py
```diff
@@ -0,0 +1,142 @@
...
+        ifgram = []
+        flag = True
+        for j in range(n):
+            ifgram.append('{}_{}'.format(inps.SLC_list[i+j],inps.SLC_list[i+j+1]))
+        ifgram.append('{}_{}'.format(SLC_list[i],SLC_list[i+n]))
+        for ifgram_name in ifgram:
+            if ifgram_name not in inps.date12_list_all:
+                # print('Interferogram not found: ', ifgram_name )
+                flag = False
+        if flag:
```
hfattahi, line 41: I'm trying to understand the logic here. It looks like for the next n interferogram if only one of the pairs is absent we won't use that list of interferograms. is this because we alway want interferograms 1-2, 2-3, 3-4, ... , n n-1 and 1-n and if one of those don't exists we can not compute the loop and therefore we drop them. is that correct?
what happens if we would have 1-2, 2-4, 4-5, ... n n-1. We would not make that closure phase?

mintpy/createmask_cp.py
```diff
@@ -0,0 +1,142 @@
...
+    cp_idx = np.array(cp_idx, np.int16)
+    cp_idx = np.unique(cp_idx, axis = 0)
+    
+    num_cp = len(cp_idx)
+    print('Number of closure measurements expected, ', len(SLC_list)-n)
+    print('Number of closure measurements found, ', num_cp)
+    
+    if num_cp <1:
+        print('No closure phase measurements found, abort')
+        sys.exit()
```
hfattahi, line 53: not a big deal but it's preferred to "raise exception" instead of sys.exit().
```suggestion
        raise Exception(No triplets found!)
```

mintpy/createmask_cp.py
```diff
@@ -0,0 +1,142 @@
...
+    phase = readfile.read(inps.ifgram_stack, box=box,print_msg=False)[0]
+    cum_cp = np.zeros((box_length, box_width), np.complex64)
+    for i in range(num_cp):
+        cp0_w = np.zeros ((box_length, box_width), np.float32)
+        for j in range(n):
+                    idx = cp_idx[i,j]
+                    cp0_w = cp0_w + phase[idx,:,:] - inps.ref_phase[idx]
+        idx = cp_idx[i,n]
+        cp0_w = cp0_w - (phase[idx,:,:]-inps.ref_phase[idx])
+        cum_cp = cum_cp + (np.exp(1j*cp0_w))
```
hfattahi, line 66: I would like to understand what is going on here. 
What is the exact quantity that we want to compute here? Isn't it the sum of the wrapped closure phases? 
It looks like you do compute the closure phases from unwrapped phases first and then you sum the complex values. I'm not sure if the angle of the sum of the complex values could be called the cumulative closure phase. 

If you really want the sum of the closure phases I assume you could keep adding the unwrapped closure phases and then at the end you could wrap it to get rid of unwrapping errors contribution to closure phase. 

If no that is not what you need would you please clarify and put in some comments describing this quantity (i.e., cum_cp).

mintpy/createmask_cp.py
```diff
@@ -0,0 +1,142 @@
...
+    # What is a good threadshod?
+    # Assume that it's pure noise so that the phase is uniform distributed from -pi to pi.
+    # The standard deviation of phase in each loop is pi/sqrt(3) (technically should be smaller because when forming loops there should be a reduction in phase variance)
+    # The standard deviation of phase in cumulative wrapped closure phase is pi/sqrt(3)/sqrt(numcp) -- again another simplification assuming no correlation.
+    # We use 2\delta as threashold -- 95.4% confidence
+
+    threashold_cp = np.pi/np.sqrt(3)/np.sqrt(numcp)*3 # 3/sigma, 99.7% confidence
+    mask = np.ones([length,width],np.float32)
+    mask[np.abs(np.angle(closurephase))>threashold_cp] = 0 # this masks areas with potential bias
+    mask[np.abs(np.abs(closurephase)/numcp < 0.3)] = 1 # this unmasks areas with low correlation (where it's hard to know wheter there is bias either)
```
hfattahi, line 128: How about exposing the magic threshold of 0.3 to the users and default it to 0.3 instead of hardcoding it.

hfattahi (commented):
Thank you @yjzhenglamarmota for this interesting PR. This is going to be a very useful addition to mintpy. I have put in some comments mainly to clarify different things. It would be also nice to update the description of the PR with some example figures of the outputs.

mintpy/closure_phase_bias.py
```diff
@@ -0,0 +1,168 @@
...
+from mintpy.objects import ifgramStack
+from mintpy.utils import readfile, writefile, ptime, utils as ut
+from mintpy import ifgram_inversion as ifginv
+
+
+
+################################################################################
+REFERENCE = """reference:
+  Zheng, Y., et al., (2022) On the Closure Phase ...
+"""
```
yunjunz, line 27: Could you add your paper info here?

mintpy/createmask_cp.py
hfattahi: ok sounds good and consistent with the manuscript.

mintpy/closure_phase_bias.py
```diff
@@ -0,0 +1,164 @@
...
+EXAMPLE = """example:
+  closure_phase_bias.py -i inputs/ifgramStack.h5 --nl 20
+"""
+
+def create_parser():
+    parser = argparse.ArgumentParser(description = 'Create an indication map for closure phase bias.')
+    parser.add_argument('-i','--ifgramstack',type = str, dest = 'ifgram_stack',help = 'interferogram stack file that contains the unwrapped phases')
+    parser.add_argument('--nl', dest = 'nl', type = int, default = 20, help = 'connection level that we are correcting to (or consider as no bias)')
+    parser.add_argument('--theta',dest = 'theta', type = float, help = 'Threashold for phase (unit: radian), default to be 3 sigma')
```
hfattahi, line 30: Would you please complete the help for this parameter. In particular "3 sigma of what distribution"? 
I know you have a nice comment written [here](https://github.com/insarlab/MintPy/pull/740/files#diff-8c628b98ef4d92190283a31ccf1098be186400827ef61e2deeabb1ffa3c07c17R137). Thank you for that. However, it would be nice if the user would have a better idea of this threshold from help message.

hfattahi (approved):
Thanks @yjzhenglamarmota for addressing my comments. It looks good to me.

mirzaees (approved):

mintpy/closure_phase_bias.py
yjzhenglamarmota: Thanks for the suggestion. I have changed the input from a phase value (unit radian), to a number of sigma (number of standard deviation for a Gaussian distribution). I have updated the help message and the code accordingly.

yunjunz (approved):
Looks good to me. Thank you @yjzhenglamarmota.

Below is my test on Sierra Negra volcano in Galapagos ([zenodo](https://zenodo.org/record/4743058)):

Screen Shot 2022-03-09 at 8 52 45 PM

PR #746, opened 2022-02-14
bugfix: timeseries2velocity `--save-res` excluded dates

### LINKED ISSUE #739: Problem with timeseries2velocity when excluding dates and outputting residuals
**Description of the problem**

When running `timeseries2velocity.py` with both `--save-res` and `--exclude` parameters, we have an issue with the resulting residuals output. 

At [L.483](https://github.com/insarlab/MintPy/blob/b33fefc4479c08448289190d586a628d2c60ffa8/mintpy/timeseries2velocity.py#L483) we layout an hdf file for saving the residuals, however we use the original timeseries file as the reference file. This creates an hdf file with the full number of dates, not taking into account the excluded dates. At [L.700](https://github.com/insarlab/MintPy/blob/b33fefc4479c08448289190d586a628d2c60ffa8/mintpy/timeseries2velocity.py#L700) when we write to this file, we only write residuals for the dates we haven't excluded, leaving the final N date entries empty (where N is the number of dates we have excluded). This means that the residuals no longer line up with the correct dates. 

I believe a straightforward solution to this would be to pass the shape of the timeseries with the excluded dates removed to `layout_hdf5` via the `ds_name_dict` argument, rather than just passing the `ref_file`. I think we'd also have to pass the metadata from the original timeseries file, although I'm not sure if I'm missing some aspect of the data structure here.

yuankailiu:
A bugfix for issue #739, which was brought up by @olliestephenson.

When saving time-series residuals during `timeseries2velocity.py`, the code did not ignore the excluded dates. It writes residuals for all the dates we haven't excluded, leaving the final N date entries empty (where N is the number of dates we have excluded). This means that the residuals no longer line up with the correct dates.

Now the code instantiates the output time-series residual hdf5 file with the correct number of dates, accounting for the excluded dates.

This bugfix is a realization of the suggested solution from @olliestephenson, in which we pass the shape of the timeseries with the excluded dates removed to `layout_hdf5` via the `ds_name_dict` argument.

Co-authored by @yunjunz 

**Reminders**

- [x] Fix #739 

### DIFF
--- a/mintpy/timeseries2velocity.py
+++ b/mintpy/timeseries2velocity.py
@@ -480,8 +480,13 @@ def run_timeseries2time_func(inps):
         for key in ['REF_DATE']:
             if key in atrR.keys():
                 atrR.pop(key)
-        writefile.layout_hdf5(inps.res_file, metadata=atrR, ref_file=inps.timeseries_file)
-
+        date_len = len(inps.dateList[0])
+        ds_name_dict = {
+            "date"       : [np.dtype(f'S{date_len}'), (num_date,), np.array(inps.dateList, dtype=np.string_)],
+            "timeseries" : [np.float32,               (num_date, length, width), None]
+        }
+        writefile.layout_hdf5(inps.res_file, ds_name_dict=ds_name_dict, metadata=atrR)
+    
 
     ## estimation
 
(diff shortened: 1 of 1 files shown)


yunjunz (approved):
Thank you @yuankailiu. LGTM!

PR #750, opened 2022-03-06
ifg_inv: bugfix for design matrix w/ date1 > date2 & release memory occupied by covariance matrix

mirzaees:

1. There is a bug in the calculation of design matrix for ifgram inversion using minimum norm phase velocity. Currently the matrix only works for a network with sequential interferograms where the secondary images are always after reference date. So basically it does not work for some networks, for example a single reference network with the reference image in the middle. Incorrect timeseries are resulted where they are zero for dates before reference. Also weighting does not work in that case.

Proposed changes:
- Correct design matrix calculation
- Correct reference date based on interferograms
- Remove the condition where rules out weighted inversion for non redundant network 

2. The second issue is regarding the memory and failure of Dask during inversion. The default of ifgram inversion is to not calculate/write covariance matrix of the timeseries which is a huge 4D matrix. But the problem is that it allocates memory for covariance matrix during inversion even when you don't want to calculate it

proposed changes:
- Release memory occupied by covariance matrix if it is not opted to calculate for.

### DIFF
--- a/mintpy/ifgram_inversion.py
+++ b/mintpy/ifgram_inversion.py
@@ -350,6 +350,7 @@ def estimate_timeseries(A, B, y, tbase_diff, weight_sqrt=None, min_norm_velocity
                 num_inv_obs       - 1D np.ndarray in size of (num_pixel), number of observations (ifgrams / offsets)
                                     used during the inversion
     """
+
     y = y.reshape(A.shape[0], -1)
     if weight_sqrt is not None:
         weight_sqrt = weight_sqrt.reshape(A.shape[0], -1)
@@ -372,11 +373,11 @@ def estimate_timeseries(A, B, y, tbase_diff, weight_sqrt=None, min_norm_velocity
         return ts, inv_quality, num_inv_obs
 
     # check 2 - matrix invertability (for WLS only because OLS contains it already)
-    if weight_sqrt is not None:
-        try:
-            linalg.inv(np.dot(B.T, B))
-        except linalg.LinAlgError:
-            return ts, inv_quality, num_inv_obs
+    #if weight_sqrt is not None:
+    #    try:
+    #        linalg.inv(np.dot(B.T, B))
+    #    except linalg.LinAlgError:
+    #        return ts, inv_quality, num_inv_obs
 
     ##### invert time-series
     try:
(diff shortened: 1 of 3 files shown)


mintpy/objects/cluster.py
```diff
@@ -328,9 +329,10 @@ def collect_result(self, futures, results, box, submission_time):
                     results[i][y0:y1, x0:x1] = sub_result
 
                 else:
-                    msg = "worker result has unexpected dimension: {}".format(num_dim)
-                    msg += '\nit should be either 2 or 3 or 4!'
-                    raise Exception(msg)
+                    if i != 1:
+                        msg = "worker result has unexpected dimension: {}".format(num_dim)
+                        msg += '\nit should be either 2 or 3 or 4!'
+                        raise Exception(msg)
```
yunjunz, line 335: I don't get the rationale behind this change. Could you explain or better, comment in the code?

mintpy/ifgram_inversion.py
```diff
@@ -372,11 +373,11 @@ def estimate_timeseries(A, B, y, tbase_diff, weight_sqrt=None, min_norm_velocity
...
-    if weight_sqrt is not None:
-        try:
-            linalg.inv(np.dot(B.T, B))
-        except linalg.LinAlgError:
-            return ts, inv_quality, num_inv_obs
+    #if weight_sqrt is not None:
+    #    try:
+    #        linalg.inv(np.dot(B.T, B))
+    #    except linalg.LinAlgError:
+    #        return ts, inv_quality, num_inv_obs
```
yunjunz, line 380: I remember our offline communication about this. Could you please add the comments here in the code also, so that we could keep track of the reasoning behind this change?

yunjunz:
Thank you @mirzaees for the fixing, it looks great! I only have two minor comments above that I hope you could add to the code.

mirzaees:
@yunjunz Thanks for reviewing, your requested changes are committed

mintpy/objects/cluster.py
hfattahi: @mirzaees 
I'm also trying to understand this part. Do you mean that when i==1, that is when the sub_result represents covariance matrix?
Is that because the ```func``` that we pass in, has always a form of :

X, covX, ... = func(some inputs)

?

mirzaees: @hfattahi and @yunjunz 
You are right, the logic is not always valid
It is correct for ifgram inversion but not other places, dem_erro.py is also using dask with different outputs. I will modify this in a way that 'ifgram_inversion_patch' function does not return cov matrix if calc_cov is False

hfattahi (commented):
Thanks @mirzaees it looks good to me as far as I can tell. We don't have unit tests in place and I have not run your PR on any dataset. Since the PR touches core inversion module, before approving, could you or @yunjunz test and confirm that the regular workflows are not broken anywhere.

mirzaees:
@hfattahi Thank you, I have tested before committing and it works for me but it would be great if @yunjunz can test as well.

mintpy/objects/cluster.py
yunjunz: I adjusted the code to check None in the output so that the DaskCluster object could support None also.

yunjunz (approved):
Looks all good to me. Thank you @mirzaees.

PR #751, opened 2022-03-16
image_stitch - bugfix for bool / integer matrix

olliestephenson:
Currently `image_stitch.py` can't correctly merge mask files. It puts the merged arrays into an array initialised as `np.nan`. Nans evaluate to `True`, meaning that we are left with blocks outside the data region that are set to `True`. Adding a simple check to see if we're merging boolean arrays, then setting the default values to `False` seems to fix the issue. Am I missing a possible bug here?


yunjunz (approved):
Good catch @olliestephenson. This is a bug, for both boolean and integer types of the input matrix. Thank you!

PR #754, opened 2022-03-19
get_design_matrix4time_func():  bugfix for `poly_order = 0`

zhurewitz:
Resolves a bug which creates design matrices with columns of zeros


yunjunz (approved):
LGTM, thank you @zhurewitz for the bug fix!

PR #774, opened 2022-05-21
use `sh` instead of `bash` in `load_data_aoi.sh`

avalentino:
**Use the correct shell in the shebang string**

The `mintpy/sh/load_data_aoi.sh` script is supposed to work with a standard unix `sh` shell but it actually contain some statements using a `bash` specific syntax.

```
$ sh -n mintpy/sh/load_data_aoi.sh 
mintpy/sh/load_data_aoi.sh: 49: Syntax error: redirection unexpected
```

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/mintpy/sh/load_data_aoi.sh
+++ b/mintpy/sh/load_data_aoi.sh
@@ -1,4 +1,4 @@
-#! /bin/sh
+#! /bin/bash
 # Extract input datasets in mintpy/inputs for an area of interest (AOI) from the one of the whole frame
 # Parameters: SNWE : string for the AOI
 #             step : string/number, output resolution in degree


avalentino:
Just for my curiosity, what is the purpose of having the `mintpy/sh/*.sh` scripts inside the `mintpy` Python package folder?
It seems to me that none of them is used in the Python code, so probably it could be a good idea to move them to another place.
Do I miss something?

yunjunz:
You are right. Only `sh/plot_smallbaselineApp.sh` was used before in the code, but not anymore. I like the idea of moving them to another place. Do you have any suggestions for it?

yunjunz (approved):
LGTM.

avalentino:
> You are right. Only  'sh/plot_smallbaselineApp.sh` was used before in the code, but not anymore. I like the idea of moving them to another place. Do you have any suggestions for it?

Probably it could be a `scripts` folder in the project root.
If you need to install those scripts, you could use the `scripts` argument  of `setuptools.setup`.
In that case, anyway, I recommend to prefix your scripts with e.g. `mintpy-` in order to make it easy to recognize mintpy related stuff in the global `/usr/bin/` folder (on unix at least).

If you want I can prepare a patch for it, but it will be not before a week.
I will be traveling the next week.

yunjunz:
Having them in a `scripts` folder sounds good to me. Since they are not used currently, we don't need to install them. One could easily find them in the repo if they want to use them independently. I will wait for your PR then.

PR #813, opened 2022-07-14
readfile: fix gdal_read to accept Paths

scottstanie:
GDALs python bindings fail for Path objects

https://github.com/OSGeo/gdal/issues/1613

The `os.fspath` was recommended in the issue over just using `str()` to convert.

Without altering a `Path` input, right now GDAL throws `RuntimeError: not a string`. This uses `os.fspath` to convert the Path to something GDAL will accept.

**Reminders**

### DIFF
--- a/mintpy/utils/readfile.py
+++ b/mintpy/utils/readfile.py
@@ -1458,6 +1458,8 @@ def read_gdal_vrt(fname):
         raise ImportError('Cannot import gdal and osr!')
 
     # read dataset using gdal
+    # Using os.fspath to convert Path objects to str, recommended by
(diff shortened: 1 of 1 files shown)


jhkennedy:
There's some additional context in the GDAL issue, namely, why `os.fspath` should be preferred over `str` which is the more common approach I've seen:
https://github.com/OSGeo/gdal/issues/1613#issuecomment-824703596

yunjunz:
Thank you @scottstanie for the fix. Could you add the comment you and Joseph have in the code as well?

yunjunz (approved):

PR #821, opened 2022-07-27
return empty in `main()` to have error code of 0 for normal complete

bbuzz31:
Simply remove the unused paths that are returned by `prep_aria.main()` per https://github.com/aria-tools/ARIA-tools/pull/313


yunjunz (approved):
LGTM. Thank you for bringing this up @bbuzz31.

PR #827, opened 2022-08-11
automate PyPI publishing with GitHub Actions

yunjunz:

+ version: replace version_num with version for easy re-use, and update its usage in __inti__.py

+ setup: use sys.path.append() to replace open() to grab version from version.py

+ add .github/workflows/publish-to-test-pypi.yml:
   - triggered for push event on main branch
   - run only if repo owner is insarlab or yunjunz, due to the limited access to the PyPI token
   - use "pip install" and "python -m build" to build the binary wheel and source tarball
   - use pypa/gh-action-pypi-publish action to push to PyPI and TestPyPI
   - enable skip_existing for TestPyPI to avoid frequent existing errors
   - push to PyPI only for released version and when owner is insarlab

+ repo setting: add github secrets from pypi and test-pypi, to support the above action

**Reminders**

- [x] Pass [Circle CI test](https://app.circleci.com/pipelines/github/yunjunz/MintPy/1389/workflows/246eb9cf-cb88-48ed-a9fd-5c2bd1aed041/jobs/1391) (green)

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yml
+++ b/.github/workflows/publish-to-test-pypi.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build-n-publish:
-    if: github.repository_owner == 'insarlab' || github.repository_owner == 'yunjunz'
+    if: github.repository_owner == 'insarlab'
 
     name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
     runs-on: ubuntu-latest
@@ -46,11 +46,11 @@ jobs:
       with:
         password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
-        skip_existing: true
+        skip_existing: false
         verbose: true
 
     - name: Publish released version 📦 to PyPI
-      if: startsWith(github.ref, 'refs/tags/v') && github.repository_owner == 'insarlab'
+      if: startsWith(github.ref, 'refs/tags/v')
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/publish-to-test-pypi.yml
+++ b/.github/workflows/publish-to-test-pypi.yml
@@ -1,0 +1,56 @@
+# link: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+name: publish distributions 📦 to PyPI and TestPyPI
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-n-publish:
+    if: github.repository_owner == 'insarlab' || github.repository_owner == 'yunjunz'
+
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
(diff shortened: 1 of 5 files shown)


yunjunz:
@jhkennedy and @avalentino could you take a look at this PR when you got a chance? I appreciate any feedback. The auto-publishing is triggered only after the PR is merged, so it's not shown here, you could check the github action output [here](https://github.com/yunjunz/MintPy/runs/7796168927?check_suite_focus=true) on my fork.

.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,56 @@
...
+      if: github.ref == 'refs/heads/main'
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+        skip_existing: true
+        verbose: true
+
+    - name: Publish released version 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags/v') && github.repository_owner == 'insarlab'
```
jhkennedy, line 51: Right now, since you just have:
```yaml
on:
  push:
    branches:
      - main
```

I don't thing this will run as as the [`github.ref` context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) will always be the branch one like `refs/heads/main`

If you change the trigger to:
```yaml
on:
  push:
    branches:
      - main
    tags:
      - v*
```

you'll also get `refs/tags/v...`

.github/workflows/publish-to-test-pypi.yml
```diff
@@ -0,0 +1,56 @@
...
+name: publish distributions 📦 to PyPI and TestPyPI
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-n-publish:
+    if: github.repository_owner == 'insarlab' || github.repository_owner == 'yunjunz'
```
jhkennedy, line 11: I'm not sure this line is necessary -- just trying to protect forks?

yunjunz: yes, this is trying to skip running for push events that occurred on forks, since the publishing to PyPI from forks will for sure fail due to the lack of PyPI token, and it should be. This is different from the docker building, which is perfectly fine for every fork to have its own docker image.

.github/workflows/publish-to-test-pypi.yml
yunjunz: Thanks! I added your suggested change.

.github/workflows/publish-to-test-pypi.yml
jhkennedy: :+1: I'd probably just let it fail in forks since you can turn it on/off in setting but this is likely better.

jhkennedy (approved):

setup.py
```diff
@@ -5,26 +5,20 @@
...
-
+import os
+import sys
 # Always prefer setuptools over distutils
 from setuptools import setup, find_packages
 
+# Grab version and description from version.py
+# link: https://stackoverflow.com/questions/53648900
+sys.path.append(os.path.dirname(__file__))
+from mintpy.version import version, description
```
avalentino, line 16: Instead of using this trick you could move to a `setup.cfg` based system (https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), which is declarative and cleaner IMHO.

setup.py
```diff
@@ -5,26 +5,20 @@
...
 
-# Grab from README file: long_description
+# Grab long_description from README.md
 with open("docs/README.md", "r") as f:
     long_description = f.read()
```
avalentino, line 21: `setup.cfg` would help also i this case

avalentino (approved):
LGTM

setup.py
jhkennedy: [`setuptools_scm`](https://github.com/pypa/setuptools_scm) also handles all the versioning work being done here implicitly. Only caveat is you need to make sure the git tags have also been fetched at package build time

yunjunz: Thanks, `setup.cfg` makes all the sense! I will try to convert it in another PR.

PR #828, opened 2022-08-13
lazy import via `mintpy.cli` for more responsive CLI

avalentino:

The CLI stuff is move to dedicated modules in the `mintpy.cli` sub-package.
In this way it is possible to define the single-entry point argument parser without importing all the processing modules.
Processing modules are imported only when necessary depending on the selected sub-command.

See also https://github.com/insarlab/MintPy/pull/823#issuecomment-1204559435.

**Reminders**


yunjunz:
Thank you @avalentino for the useful and large PR! A quick test shows that the `mintpy` response time is significantly reduced.

I am currently occupied by the UNAVCO ISCE+ short course preparation, will look at this PR more carefully after that, probably next week.

setup.py
```diff
@@ -80,71 +91,71 @@
...
+            'info.py = mintpy.cli.info:main',
+            'iono_tec.py = mintpy.cli.iono_tec:main',
+            'load_data.py = mintpy.cli.load_data:main',
+            'load_gbis.py = mintpy.cli.load_gbis:main',
+            'local_oscilator_drift.py = mintpy.cli.local_oscilator_drift:main',
+            'lookup_geo2radar.py = mintpy.cli.lookup_geo2radar:main',
+            'mask.py = mintpy.cli.mask:main',
+            'modify_network.py = mintpy.cli.modify_network:main',
+            'multilook.py = mintpy.cli.multilook:main',
             'multi_transect.py = mintpy.multi_transect:main',
```
jhkennedy, line 115: Why's this one the only untouched one?

setup.py
```diff
@@ -70,8 +70,19 @@
...
+        "solidearth": ["pysolid"],
+        "gps": ["proj", "dateutil"],
+        "progress": ["tqdm"],
+        "isce": ["isce"],
+        "legacy": ["pyaps"],
+        "extra": ["gdal", "pillow"],
+        "solvers": ["mosec"],
+        "gbis": ["geoid"],
+        "kite": ["kite"],
+        "fractal": ["pyfftw"],
```
jhkennedy, line 85: With this many extras, it might be worth adding an `all` target that have all the extra requirements in it (except `platemotion` maybe). 

Though, TBF I'll likely never spell out any of the extras as I'll build a conda env first and then pip install into it with `--no-deps`.

mintpy/__main__.py
```diff
@@ -58,452 +58,452 @@ def _autocomplete(parser):
 # in alphabetical order
 # A-G
 def get_add_parser(subparsers=None):
-    from . import add
+    from .cli import add
```
jhkennedy, line 61: Any reason we're preferring relative imports over absolute? Generally the python community/best practices prefer absolute as it makes the package namespace a lot clearer, but does require a little more work when refactoring (though any good IDE will handle all that work for you).

jhkennedy (commented):
@yunjunz I gave this a look over the weekend and overall it looks exactly as I suspected it would -- overall looks great to me.

setup.py
avalentino: Multi-transect uses the old getopt module instead of argparse so it is not possible to integrate it in the current "single-entrypoint" without a porting.
I could try to doit but I was not sure if `multi_transect` is actually used and updated or something legacy.
Please let me know.

setup.py
avalentino: Good idea!
Done.
I also left out "legacy".

mintpy/__main__.py
avalentino: no, just my habit.
I will change all imports to absolute ...

avalentino: Done!

setup.py
yunjunz: `multi_transect` is a legacy script and has not been actively maintained. Keeping it out of this PR is perfectly fine with me.

yunjunz (approved):
Thank you @avalentino for the PR and @jhkennedy for the review! I really like the changes here. All the command line helps are now significantly speeded up, which improves the user experience a lot. Cheers!

avalentino:
@yunjunz there is a small issue.
In `mintpy.cli.prep_gamma` there is 
```python
from mintpy.objects.sensor import SENSOR_NAMES
```
that triggers the import of numpy and h5py in `mintpy.objects.giant`.

You chan check it using the following command:
```
python3 -X importtime -m mintpy -h
```

My original solution was to have a local copy of the `SENSOR_NAMES` list, that, I agree, is not nice.
Another option could be to move `SENSOR_NAMES` to another module that can be imported without triggering the loading of heavy-waight modules like h5py and numpy.
How do you propose to proceed?

yunjunz:
Thank you for the `importtime` command, that's very helpful. The triggering of numpy import is because of the non-empty `mintpy.objects.__init__.py` file (this should be changed in the next major version), but modifying this now would require quite a lot of changes. How about moving `sensor.py` from `mintpy.objects` to `mintpy.utils` sub-module?

The related changes would be only a few places.

avalentino:
> Update: rethinking about this makes me realize that making a copy as you did before was a pretty good solution! Shall we revert back to that?

Probably it is the easiest solution.
If you already plan a refactoring, then it should not be a bug issue.

avalentino:
OK, @yunjunz 
Commit done.

avalentino:
Thanks @yunjunz

yunjunz:
Thank you @avalentino for the PR!

PR #841, opened 2022-08-29
readfile: allow ENVI files created with SUFFIX=ADD

scottstanie:

The GDAL creation option "SUFFIX=ADD" will append .hdr to the name, rather than replace the suffix. https://gdal.org/drivers/raster/envi.html

This addition will search for that extra naming convention for ENVI files and not miss the metadata.

**Reminders**

### DIFF
--- a/mintpy/utils/readfile.py
+++ b/mintpy/utils/readfile.py
@@ -1050,6 +1050,7 @@ def read_attribute(fname, datasetName=None, metafile_ext=None):
             fname + '.xml',
             fname + '.par',
             os.path.splitext(fname)[0] + '.hdr',
+            fname + '.hdr',
             fname + '.vrt',
             fname + '.aux.xml',
         ]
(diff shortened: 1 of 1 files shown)


mintpy/utils/readfile.py
```diff
@@ -1050,6 +1050,7 @@ def get_hdf5_dataset(name, obj):
             fname + '.xml',
             fname + '.par',
             os.path.splitext(fname)[0] + '.hdr',
+            fname + '.hdr',
```
yunjunz, line 1053: Given that the one created with `SUFFIX=ADD` is more unique than the one created with `SUFFIX=REPLACE`, shall we put this in front of the `SUFFIX=REPLACE` version above?

yunjunz: It would be good to write comments after these two lines for the `SUFFIX=ADD/REPLACE` option, to easily remind ourselves in the future about this change.

yunjunz (approved):
Thank you @scottstanie.

PR #842, opened 2022-09-04
unmask ref pixel in ts2vel and tsview

bbuzz31:
when the `timeseriesResidual.h5` is created with `timeseries2velocity.py --save-res`, the reference pixel time series is overwritten with `np.nan`. 

Then, when plotting (e.g., `tsview.py`) the whole dataset is masked out because it gets referenced back to the nan time series. This fix just preserves the 0 vector timeseries at the reference point.

**Reminders**


yunjunz (approved):
LGTM. Thank you @bbuzz31.

PR #868, opened 2022-10-13
`plate_motion` refactoring: remove pkg dependency

yuankailiu:

+ Use matrix rotation of the cartesian frame to compute the plate motion from the Euler pole

+ Add reference to Pichon et al., 1973, Cox et al., 1986, ESA webpage

+ thus we can remove the dependency on (1) astropy, and  (2) platemotion: https://github.com/lcx366/PlateTectonic

**Reminders**


yunjunz:
Thank you @yuankailiu for this PR, it's very exciting! Removing the extra dependency is critical for us to integrate it into the routine workflow.

After looking at it, it makes a lot of sense to create an object/class for the Euler pole initiation/calculation, maybe in an independent script, e.g. `objects/euler_pole.py`, in a similar form as https://github.com/joferkington/euler_pole/blob/master/euler_pole/euler_pole.py. We could try it in the coming week.

It's also a good time to include a basic unit test to compare the calculated PMM at a few selected lat/lon using mintpy with the one using the unavco web calculator (https://www.unavco.org/software/geodetic-utilities/plate-motion-calculator/plate-motion-calculator.html). After the above is finished I assume.

yuankailiu:
Hi @yunjunz, I am done with the changes. Please take a look!

yunjunz:
> Hi @yunjunz, I am done with the changes. Please take a look!

Great, I will check it now. Thank you @yuankailiu for the refactoring.

yunjunz:
Hi @yuankailiu, one last question on this PR: the calculated plate motion in the vertical direction is forced to zero in the previous version (https://github.com/insarlab/MintPy/blob/c15ad0f75ab042130f108512179d255e098a6bdd/mintpy/plate_motion.py#L133), but this is not the case for this PR. Shall we still enforce it?

yuankailiu:
> Hi @yuankailiu, one last question on this PR: the calculated plate motion in the vertical direction is forced to zero in the previous version (
> 
> https://github.com/insarlab/MintPy/blob/c15ad0f75ab042130f108512179d255e098a6bdd/mintpy/plate_motion.py#L133
> 
> ), but this is not the case for this PR. Shall we still enforce it?

I think we can keep the verticles. People will need to understand when using the non-spherical Earth, we will inevitably have some tiny verticles.

yunjunz:
> I think we can keep the verticles. People will need to understand when using the non-spherical Earth, we will inevitably have some tiny verticles.

Thank you Kai. Good point. 

Related to the vertical component, when `ellips=False`, the resulting vertical component is extremely tiny, which is expected, but the spatial pattern is strange, as shown below. Shall we set that to zero in `euler_pole.py` in that scenario?

+ San Francisco Sen DSC - Assume Ellipsoid
Ellips

+ San Francisco Sen DSC - Assume Sphere
Sphere

yuankailiu:
I see. Thanks a lot for testing this! Yes, I think in that case (when ellips=False), we can just set it to zero. Good point.

yunjunz (approved):
Thank you @yuankailiu for this PR! I really like its structure: the class definition and unit tests could serve as a template/example for future new features.

PR #874, opened 2022-10-21
`save_roipac`: support non-velocity dataset from the velocity file

bjmarfito:

- Ability to extract a specific dataset from a velocity file and save it in ROIPAC format.
- Removed behavior to make datasets into displacement values with the exception of the velocity dataset

**Reminders**

### DIFF
--- a/mintpy/save_roipac.py
+++ b/mintpy/save_roipac.py
@@ -44,7 +44,7 @@ def read_data(inps):
     # various file types
     print(f'read {inps.dset} from file {inps.file}')
     k = atr['FILE_TYPE']
-        if k == 'velocity':
+    if k == 'velocity':
         # read/prepare data
         if inps.dset == None:
             inps.dset = 'velocity'
--- a/mintpy/save_roipac.py
+++ b/mintpy/save_roipac.py
@@ -46,30 +46,29 @@ def read_data(inps):
     k = atr['FILE_TYPE']
     if k == 'velocity':
         # read/prepare data
-        if inps.dset == None:
+        if not inps.dset:
             inps.dset = 'velocity'
-            print('No selected datset, assuming chosen dataset is velocity')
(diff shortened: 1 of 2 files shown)


mintpy/save_roipac.py
```diff
@@ -46,13 +46,16 @@ def read_data(inps):
...
 
         # velocity to displacement
-        print('convert velocity to displacement for {}'.format(atr['DATE12']))
-        date1, date2 = atr['DATE12'].split('_')
-        dt1, dt2 = ptime.date_list2vector([date1, date2])[0]
-        data *= (dt2 - dt1).days / 365.25
+        #print('convert velocity to displacement for {}'.format(atr['DATE12']))
+        #date1, date2 = atr['DATE12'].split('_')
+        #dt1, dt2 = ptime.date_list2vector([date1, date2])[0]
+        #data *= (dt2 - dt1).days / 365.25
```
yunjunz, line 59: What's the rationale to comment out the conversion part? Quite some modeling codes expect the old ROI_PAC file, which does not have a velocity format. Thus, converting it into displacement is needed.

If your goal is saving the other datasets inside the velocity file, such as the step function, seasonal fitting etc., how about adding an if check here, to keep the old behavior for `velocity` dataset only?

bjmarfito: Hi @yunjunz, sorry for not providing the rationale for commenting out the lines of code. Yeah, you're right I was thinking that for the specific velocity dataset such as step, functional, and seasonal fitting to not convert it to displacement. Yeah, I forgot to consider the velocity dataset for this will need conversion to displacement. 

I will edit this PR based on your suggestion.

yunjunz (approved):
Looks all good to me now. Thank you @bjmarfito for contributing!

bjmarfito:
You're welcome and thank you @yunjunz!

PR #875, opened 2022-10-22
`dem_error`: change default from phase to phase velocity

mirzaees:

In this PR, the DEM error calculation default is changed from phase history to phase velocity history based on the conclusion of the paper proposing the algorithm and experiment of geolocation correction on urban area


yunjunz (approved):
LGTM. Thank you @mirzaees.

PR #886, opened 2022-11-03
docs/install: set env variables for more robust handling of HDF5 and GDAL VRT

mohseniaref:

**modify code in order to prevent being locked while reading hdf5 file**

I faced the following error while modifying network “OSError: Unable to open file ”. The solution which worked for me was preventing hdf5 file being locked by command which was added to script


yunjunz (approved):
Thanks @mohseniaref for reporting the issue. Since this issue could happen to other steps as well, the best practice would be setup the environment variables in our source files. So I go ahead and update the installation note for this; and add two more based on my previous experience. Cheers.

Issue #892, opened 2022-11-07
abnormal large residual from `dem_error` for `timeseries_rms`


yunjunz:
With the default minization in `dem_error.py` switched from `phase` to `phase velocity` in #875, I noticed the time-series residual after `dem_error` got much bigger for one dataset I have, as shown below. Although this change of behavior is reasonable given their different objective functions, they are not suitable anymore for the following `timeseries_rms` step in `smallbaselineApp`. 

The right way would be: keep phase velocity as default in `correct_topography`, move the `residual_RMS` step after the time function fitting in the `velocity` step, which has more powerful / advanced time func capability; then we won't need the TS residual from `dem_error` anymore. Before this change, I would like to revert #875 to minimize its impact on `timeseries_rms`.

@mirzaees @hfattahi let me know what do you think please.

+ Residual time-series with **phase** minimization

<img width="1612" alt="ts_res_pha" src="https://user-images.githubusercontent.com/13143710/200223995-c63d9c29-d955-4a5f-8eed-02d500138154.png">

+ Residual time-series with **phase velocity** minimization

<img width="1612" alt="ts_res_phaVelo_v1" src="https://user-images.githubusercontent.com/13143710/200224074-75f2e38e-c05a-4469-ab64-946072285d99.png">

falkamelung:
The one example we looked at in detail (Miami full-res) you absolutely need phase velocity. And isn't that what we would expect? I have to look at the paper again, but don't Fattahi&Amelung 2013 make the case that phase velocity needs to be used?   Can you remind us what was the rationale to use phase minimization  in MintPy?

yunjunz:
One could always manually turn ON the phase velocity whenever they need it. My purpose here is to not disturb other steps outside of `correct_topography`.

mirzaees:
Hi @yunjunz!
After spending some time on this to understand why it is happening, I believe there might be another bug in the calculation of design matrix (G_defo):
https://github.com/insarlab/MintPy/blob/64ffdefa14cdff5c1024bb522a7f6f1e79de1103/mintpy/utils/time_func.py#L285-L286
I checked it for polynomial model (you might want to check other deformation models as well), I think it should be like:

```
    for i in range(1, degree+2):
        A[:,i-1] = (yr_diff**i) / np.math.factorial(i)
```

Please check it if that resolves the problem

yunjunz:
Thank you @mirzaees for looking into this. For a polynomial with an order of 3, the math form should be:

```shell
[...]

PR #905, opened 2022-11-11
`dem_error` w/ phase velocity: rm all-zero column in design matrix

yunjunz:

+ `dem_error.py --phase-velocity`: remove the all-zero column in the design matrix `G` and the all-one column in the design matrix `G0`, as suggested by @mirzaees in #892. This is because the constant intercept term of the polynomial is not estimated in the phase velocity minimization scenario. Including this all-zero column in `G` could sometimes introduce abnormally large eigenvalues in the least-squares inversion, resulting in an unrealistic TS residual.

**Reminders**


mirzaees (approved):
Thank you @yunjunz! It looks good.
The only point that we should not forget is that timeseries residuals in line 266 lacks a constant if using phase velocity and results large residuals, hence not reliable.
It will be resolved after moving this part to velocity estimation

PR #922, opened 2022-11-30
closure_phase_bias: use `wrapPhase` if available

yjzhenglamarmota:
Due to some issues with filtering implemented in ISCE2, I added the option of using wrapped phase (read from fine.int) to compute closure phase.
Now if there are wrapped phase in the ifgramStack.h5, wrapped phase will be used instead of the unwrapped phase.

+ docs/install: bugfix in #923 for the order of `conda activate mintpy` and `mamba install isce2`

### DIFF
--- a/src/mintpy/objects/stack.py
+++ b/src/mintpy/objects/stack.py
@@ -1169,7 +1169,7 @@ class ifgramStack:
             phase = self.read(datasetName='wrapPhase',box=box, print_msg=False)
 
         else:
-            phase = self.read(datasetName='unwrapPhase',box=box, print_msg=False)
+            phase = self.read(box=box, print_msg=False)
 
         ## calculate the 3D complex seq closure phase
         cp_w = np.zeros((num_cp, np.shape(phase)[1], np.shape(phase)[2]), dtype=np.complex64)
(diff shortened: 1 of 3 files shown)


src/mintpy/objects/stack.py
```diff
@@ -1164,20 +1164,20 @@ def get_sequential_closure_phase(self, box, conn, post_proc=None):
                 raise Exception(f"No triplets found at connection level: {conn}!")
 
         ## read data
-        phase = self.read(box=box, print_msg=False)
-        ref_phase = self.get_reference_phase(dropIfgram=False)
-        for i in range(phase.shape[0]):
-            mask = phase[i] != 0.
-            phase[i][mask] -= ref_phase[i]
```
yunjunz, line 1171: If I understood it right, spatial referencing and masking are needed if using `unwrapPhase`.

yjzhenglamarmota: Spatial referencing is not needed. Since no moisture change should have a naturally zero closure phase, choosing a reference point in the wrong place can lead to misinterpretations. I tested in Central Valley and proved spatial referencing is not needed.
Not sure about the masks. Why do we need masks here in the first place?

yjzhenglamarmota:
Now added an explanation for removing the reference point when creating the wrapped closure phase.
Tests in the Brystol-Barstow area again confirmed that a reference point is not needed.

src/mintpy/objects/stack.py
yunjunz: Thank you for your explanation offline. I agree with you that spatial referencing is not needed here. Since masking is serving the referencing process, it's also not needed anymore.

yunjunz (approved):
LGTM, thank you @yjzhenglamarmota!

PR #935, opened 2022-12-23
`euler_pole`: add PMM plotting functions

yuankailiu:
+ add plate boundary files from GSRMv2.1 and MORVEL56

+ functions to read the boundary files and plot

Check here for the usage: https://github.com/yuankailiu/utils/blob/main/notebooks/PMM_plot.ipynb

**Reminders**


yuankailiu:
Checks seem to fail at testing `/root/tools/MintPy/tests/objects/ionex.py`

But I did not change this file.

yunjunz:
> Checks seem to fail at testing `/root/tools/MintPy/tests/objects/ionex.py`
> 
> But I did not change this file.

The same test works well locally on my laptop, and I could not find out the reason on circle CI. As you said, this PR did not touch this part of the code, let's ignore this error.

src/mintpy/objects/euler_pole.py
```diff
@@ -17,9 +17,15 @@
...
+import os
 
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import matplotlib.pyplot as plt
 import numpy as np
 import pyproj
+from shapely import geometry
 
 from mintpy.constants import EARTH_RADIUS
```
yunjunz, line 31: I would suggest defining a variable here for the two PMM plate boundary files, such as below, then the other code could just use `PLATE_BOUNDARY_FILE['GSRM']` as the file name, without caring about the detail.

```python
import mintpy

PLATE_BOUNDARY_FILE = {
    'GSRM' : os.path.join(mintpy.__path__, 'data/plate_boundary/GSRM/GSRM_plate_outlines.gmt'),
    'MORVEL56' : os.path.join(mintpy.__path__, 'data/plate_boundary/MORVEL56/All_boundaries'),
}
```

yunjunz:
For the NNR PMM value/tables, maybe it's easier to use if we write them in Python, similar to what we have for [ITRM2014-PMM](https://github.com/insarlab/MintPy/blob/4e88f085cf6e94d36924599fab05935f0c806b9f/src/mintpy/cli/plate_motion.py#L15)?

If so, we could move these variable definitions from `mintpy.cli.plate_motion` into `mintpy.plate_motion`, for easier usage.

src/mintpy/objects/euler_pole.py
yuankailiu: sounds good

yuankailiu:
> For the NNR PMM value/tables, maybe it's easier to use if we write them in Python, similar to what we have for [ITRM2014-PMM](https://github.com/insarlab/MintPy/blob/4e88f085cf6e94d36924599fab05935f0c806b9f/src/mintpy/cli/plate_motion.py#L15), instead of a data file, which requires extra reading?
> 
> If so, we could move these variable definitions from `mintpy.cli.plate_motion` into `mintpy.plate_motion`, for easier usage.

Yes, I can do that. Now those files have different units compared with our ITRF2014 dictionary.
The files are in the format of Lat, Lon, and the rotation rate in degrees per Ma.
I can just create dictionaries for them without converting the units. Anyway, these parameters are not supposed to be used in computing the plate motion.

yuankailiu:
Please take a look at the new updates.

> If so, we could move these variable definitions from mintpy.cli.plate_motion into mintpy.plate_motion, for easier usage.

But I don't see why we will need to put those dict() in `mintpy.plate_motion`? Also where should we put the `PLATE_BOUNDARY_FILE `? Right now, it is under the `euler_pole.py`.

src/mintpy/objects/euler_pole.py
```diff
@@ -539,169 +542,203 @@ def transform_xyz_enu(lat, lon, x=None, y=None, z=None, e=None, n=None, u=None):
...
+    # check input
+    if 'GSRM' in pmm_name:
+        pmm_name = 'GSRM'
+        pmm_dict = GSRM_V21_PMM
+        coord_order = 'lola'
+
+    elif 'MORVEL' in pmm_name:
+        pmm_name = 'MORVEL'
+        pmm_dict = NNR_MORVEL56_PMM
+        coord_order = 'lalo'
```
yuankailiu, line 561: This place should be 'lola' as well.

Because I also flip the order in the file from the original `All_boundary` in MORVEL_56.

So we can just set the default as `coord_order='lola'` and keep the file as it is (and it also makes sense since it has `.gmt` extension.).

But this way, we no longer keep the original downloaded file untouched. But anyway, I already added those `> ` signs in the file.

yunjunz (approved):
Thank you @yuankailiu for the nice PR. I went through everything and am able to produce a nice plot as below. All looks good to me, pending the above comments as we discussed offline.

```python
from matplotlib import pyplot as plt
from mintpy.objects import euler_pole
from shapely import geometry

plate_pmm = euler_pole.ITRF2014_PMM['Arabia']
epole_obj = euler_pole.EulerPole(wx=plate_pmm.omega_x, wy=plate_pmm.omega_y, wz=plate_pmm.omega_z, unit='mas/yr')

plate_boundary = geometry.Polygon(euler_pole.read_plate_outline('GSRM')['Arabia'])

fig, ax = euler_pole.plot_plate_motion(plate_boundary, epole_obj)
plt.show()
```

image

src/mintpy/objects/euler_pole.py
yunjunz: As we discussed offline:
+ let's keep the original downloaded file content untouched, for reproducibility. 
+ use `.lonlat` and `.latlon` as file extensions, instead of `.gmt`, to be more informative.
+ update the `data/plate_boundary/README.md` file to be consistent with the above changes.

yuankailiu:
Thanks @yunjunz  for testing. I finish my changes for this PR as we discussed.

By the way, now you can do the following to read the dictionary in your notebook:
```
plate_boundary = geometry.Polygon(euler_pole.read_plate_outline('GSRM', 'Arabia'))
```

yunjunz:
Thank you @yuankailiu for the changes. Built on top of yours, the commit above further simplify the `read_plate_outline()` call as below. Everything looks good to me now! Merged it when you got a chance please.

```python
plate_boundary = euler_pole.read_plate_outline('GSRM', 'Arabia')
```

PR #983, opened 2023-03-19
`prep_fringe`: load water mask into geometry file via `--water-mask`

bbuzz31:
simple change to allow watermask to be stored in geometry via prep_fringe.py.
I'm not sure how the CLI is working, but this did as expected for me.


src/mintpy/prep_fringe.py
```diff
@@ -26,6 +26,76 @@
...
+
+
+def cmd_line_parse(iargs=None):
+    parser = create_parser()
+    inps = parser.parse_args(args=iargs)
+
+    # in case meta_file is input as wildcard
+    inps.metaFile = sorted(glob.glob(inps.metaFile))[0]
+
+    return inps
```
yunjunz, line 98: This part of change seems unnecessary given the current CLI setup.

yunjunz:
The PR looks good to me now. @bbuzz31 could you try it again to confirm if it works as you expected?

bbuzz31:
@yunjunz works for me, thanks!

yunjunz (approved):

yunjunz:
> @yunjunz works for me, thanks!

Great, thanks for the PR!

PR #991, opened 2023-04-15
Added constant in ramp.py

bjmarfito:

Added estimation and removal of constant in ramp.py

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/src/mintpy/objects/ramp.py
+++ b/src/mintpy/objects/ramp.py
@@ -11,6 +11,7 @@ import numpy as np
 
 # duplicated in mintpy.cli.remove_ramp
 RAMP_LIST = [
+    'constant',
     'linear',
     'linear_range',
     'linear_azimuth',
@@ -92,6 +93,8 @@ def deramp(data, mask_in=None, ramp_type='linear', metadata=None, max_num_sample
         G = np.hstack((xx**2, xx, ones))
     elif ramp_type == 'quadratic_azimuth':
         G = np.hstack((yy**2, yy, ones))
+    elif ramp_type == 'constant'
+        G = np.hstack((ones))
     else:
         raise ValueError(f'un-recognized ramp type: {ramp_type}')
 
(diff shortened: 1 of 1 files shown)


yunjunz:
@bjmarfito could you describe your motivation for this addition? Note that since InSAR products are relative in space, we always keep the value at the reference pixel (specified with `REF_Y/X`) to zero as below, thus, will remove the effect of the estimated constant from this PR.

https://github.com/insarlab/MintPy/blob/5ca554fef324b816f9130feec567e2cf463e41d2/src/mintpy/objects/ramp.py#L110-L114

bjmarfito:
Hi @yunjunz, sorry, I missed this part of the code. My motivation for introducing this constant deramping is to estimate it using the the whole or part of the interferogram  and not just using a reference point. Please correct me if my way of thinking is wrong.

yunjunz:
> Hi @yunjunz, sorry, I missed this part of the code. My motivation for introducing this constant deramping is to estimate it using the whole or part of the interferogram and not just using a reference point. Please correct me if my way of thinking is wrong.

Hi @bjmarfito, no need to be sorry at all, I appreciate your effort of contributing back! From my understanding, the "constant deramping" is to estimate a global mean, which is intended to do spatial referencing using the global mean. I don't really recommend that personally, as it makes the interpretation complicated: there is no exact location for this spatial reference, and it could change as the dataset spatial coverage change. 

On the other hand, this is a subjective choice, it's totally fine with me if you still want it in the code, just make sure you remove the `REF_Y/X` attributes before you run this constant deramping, to turn off the re-referencing effect from the mentioned code above, in a different (maybe your own) script. I would prefer not to add this attribute removal automatically within this script for the cautious reason, so that regular users won't use it by accidence without knowing this clearly.

bjmarfito:
Hi @yunjunz, thank you for your insights. Based on your recommendations, I think it is better for me to write my personal code for this as to not confuse the other users of this package.

I will close this PR.

PR #1034, opened 2023-06-30
fix bug in gdal -> mintpy corner coordinate convention

sssangha:

Through the ARIA products workflow, or whenever loading a gdal file in general, an erroneous shift appears to be applied to the accessed geotransform and the derived `x_first` and `y_first` fields. Specifically, an extra half pixel shift (to the north and west) is applied when the gdal geotransform object already reports the coordinates of the [upper left pixel](https://gdal.org/tutorials/geotransforms_tut.html).

I've tracked this bug in a few scripts, but would like advise on where else I should look to track down potential persistent issues.

**Reminders**

### DIFF
--- a/src/mintpy/prep_aria.py
+++ b/src/mintpy/prep_aria.py
@@ -180,8 +180,8 @@ def extract_metadata(stack):
     x_step = abs(transform[1])
     y_step = abs(transform[5]) * -1.
 
-    W = transform[0] - x_step / 2.
-    N = transform[3] - y_step / 2.
+    W = transform[0]
+    N = transform[3]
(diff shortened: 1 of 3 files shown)


yunjunz (approved):
Thank you @sssangha for the careful checking! I introduced this bug in `prep_aria` in #542. As you said, gdal (and aria, hyp3, and gmtsar by extension) use the upper-left corner of the upper-left pixel as the first coordinates, which is the same as roipac and mintpy, thus, no extra half-pixel shift is needed. I believe you captured all the occurrences in the code. Feel free to squash and merge the PR whenever you are ready please.

PR #1052, opened 2023-08-01
convert coord btw. UTM and lat/lon

yunjunz:
**Description of proposed changes**

+ utils.utils0.py: add the following functions to convert coordinates between UTM and lat/lon, as described in https://github.com/insarlab/MintPy/discussions/1016#discussioncomment-6058402.
   - utm2latlon()
   - latlon2utm()
   - utm_zone2epsg_code(): bugfix if the latitude letter is not N or S
   - get_lat_lon(): convert UTM to lat/lon to ensure lat/lon output for hyp3 products. Together with #1050, this fixes #1049 

+ add `utm` to the dependency, since 1) it depends on numpy only; and 2) it's available on PyPI and conda-forge
   - requirements.txt
   - setup.py

**Reminders**

- [x] Pass Pre-commit check (green)
- [x] Pass Codacy code review (green)
- [x] Pass [Circle CI test](https://app.circleci.com/jobs/github/yunjunz/MintPy/2237) (green)
- [x] Make sure that your code follows our style. Use the other functions/files as a basis.
- [x] If modifying functionality, describe changes to function behavior and arguments in a comment below the function declaration.
- [x] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #1053, opened 2023-08-02
fix the UTM_ZONE confusion

yunjunz:

This PR fixes the previous confusion that UTM_ZONE should end with the hemisphere, not the latitude band, which is not part of UTM but part of MGRS, as described in https://github.com/insarlab/MintPy/pull/1052#discussion_r1280899939.

+ utils.utils0.utm_zone2epsg_code(): revert back the south arg feeding from UTM_ZONE

+ utils.utils0.utm2latlon(): use the `northern` arg instead of the latitude band, to match the meaning of UTM_ZONE metadata

+ utils.utils0.latlon2utm(): return east/north only, for simplicity and style consistency with utm2latlon.

+ docs/api/attributes.UTM_ZONE: add more description and examples

**Reminders**

- [x] Pass [Circle CI test](https://app.circleci.com/jobs/github/yunjunz/MintPy/2243) (green)

### DIFF
--- a/src/mintpy/utils/utils0.py
+++ b/src/mintpy/utils/utils0.py
@@ -260,17 +260,16 @@ def touch(fname_list, times=None):
 def utm_zone2epsg_code(utm_zone):
     """Convert UTM Zone string to EPSG code.
 
-    Parameters: utm_zone  - str, atr['UTM_ZONE']
+    Parameters: utm_zone  - str, atr['UTM_ZONE'], comprises a zone number
+                            and a hemisphere, e.g. 11N, 36S, etc.
     Returns:    epsg_code - str, EPSG code
     Examples:   epsg_code = utm_zone2epsg_code('11N')
     """
     from pyproj import CRS
-
-    # N is the first letter in the northern hemisphere
     crs = CRS.from_dict({
         'proj': 'utm',
         'zone': int(utm_zone[:-1]),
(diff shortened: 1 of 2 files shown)


src/mintpy/utils/utils0.py
```diff
@@ -330,11 +329,9 @@ def latlon2utm(lat, lon):
                 lon      - scalar or 1/2D np.ndarray, WGS 84 coordiantes in x direction
     Returns:    easting  - scalar or 1/2D np.ndarray, UTM    coordiantes in x direction
```
scottstanie, line 330: Should we cut out the `easting/northing` arggs from the docstring?

src/mintpy/utils/utils0.py
```diff
@@ -318,8 +317,8 @@ def utm2latlon(meta, easting, northing):
     """
```
scottstanie, line 317: likewise (i cant comment directly on the line) it looks like theres `lat, lon` as parameter descriptions here?

scottstanie (approved):

src/mintpy/utils/utils0.py
yunjunz: `easting/northing` are returned parameters here, I would prefer to keep them to help understand the function usage. 

If you meant the docstring style of input/output here is not standardized as the `compass` style, I agree. I would love to change the style in another PR for all functions.

PR #1063, opened 2023-08-08
seq. closure phase: bring back spatial referencing for ARIA products

yunjunz:

This PR brings back the spatial referencing while calculating the sequential closure phase (which was removed in #922), for ARIA products only, to avoid the abnormal closure phase calculation result, as shown below. This is happening to ARIA products only, there is negligible impact on the interferogram stack processed with ISCE-2 from SLC (as @yjzhenglamarmota and I have tested before). 

My guess is that the phase stitching applied during ARIA product preparation in ARIA-tools broke the temporal consistency of the native unwrapped phase (when used without spatial referencing). What do you think @sssangha @mgovorcin @dbekaert?

### [San Francisco Bay SenDT042](https://github.com/insarlab/MintPy/blob/main/docs/demo_dataset.md#sentinel-1-on-san-francisco-bay-with-aria) (downloaded from ARIA and processed with ARIA-tools)

+ Without spatial referencing

Screenshot 2023-08-08 at 13 55 17

+ With spatial referencing

Screenshot 2023-08-08 at 13 55 07

### [Sierra Negra SenDT128 dataset](https://zenodo.org/record/4743058) (processed with isce2/topsStack)

+ Without spatial referencing

Screenshot 2023-08-08 at 14 25 18

+ With spatial referencing

Screenshot 2023-08-08 at 14 25 27

**Reminders**

- [x] Pass [Circle CI test](https://app.circleci.com/pipelines/github/yunjunz/MintPy/2266/workflows/1a40ff8a-88a7-4d43-8c5d-2c1cc58b9239/jobs/2274) (green)

### DIFF
--- a/src/mintpy/objects/stack.py
+++ b/src/mintpy/objects/stack.py
@@ -1191,6 +1191,15 @@ class ifgramStack:
         print(f'reading {ds_name} to compute closure phases')
         phase = self.read(datasetName=ds_name, box=box, print_msg=False)
 
+        # apply spatial referencing to ARIA products
+        processor = self.metadata.get('mintpy.load.processor', 'isce')
+        if ds_name == 'unwrapPhase' and processor in ['aria']:
+            print(f'apply spatial referencing to {processor} products')
+            ref_phase = self.get_reference_phase(dropIfgram=False)
+            for i in range(phase.shape[0]):
+                mask = phase[i] != 0.
+                phase[i][mask] -= ref_phase[i]
+
         ## calculate the 3D complex seq closure phase
         cp_w = np.zeros((num_cp, box_len, box_wid), dtype=np.complex64)
         for i in range(num_cp):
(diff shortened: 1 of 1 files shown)


yjzhenglamarmota (commented):
Looks good to me. The amplitude for the ARIA test case is clearly wrong, though the phase looks surprisingly coherent. I agree it might be related to filtering that breaks the temporal consistency.

mgovorcin:
@yunjunz did you run reference_point.py before this?

 My guess is that GUNW ifgs might have different unwrapping seed point  between pairs (this will especially change during stitching process) as coregistration is not done to the same reference image. The requirement would be to do spatial referencing before any further processing on them. 

We do not do re-referencing of all ifgs to common reference point during prep_aria as that is done during reference_point step in the smallbaseline workflow. @sssangha might be good to add this to avoid potential misuse, what do you think?

yunjunz:
@mgovorcin The first figure is without running reference_point.py. 

Rethinking about this, I don't think the different seeding points or phase stitching would introduce this, because neither of them will change the wrapped phase value, which is what this figure is really using. The different reference images during pair-wise coregistration, as used in ARIA, sound more likely, as inconsistent range shifts are applied to both the resampling and SLC phase component, thus, could introduce this temporal inconsistency. Do you agree @hfattahi and @piyushrpt?

Issue #1041, opened 2023-07-06
Suggestions for MintPy version 2


falkamelung:
**Description of the desired feature**
Just want to share a few ideas. Please feel free to move this elsewhere.
- start all scripts with mp_*.  This allows to see the script names by typing `mp_ TAB` 
``` 
mp_t
mp_timeseries_rms.py         mp_tropo_gacos.py            mp_tropo_phase_elevation.py
```
(it would make it easy to combine MIntPy with Non-Mintpy scripts)

- write all executed commands in a `.commands.log` file. This would make it easy to rerun a command from the workflow using different options. And it keeps a log of manually executed commands.

- support `--subset-lalo 19.43:19.51,-155.65:-155.54` format (currently only used in *template control files)

yunjunz:
Hi @avalentino, I tried to use the argcomplete as below but was unsuccessful on the activation. I am using bash on a Mac with an M2 chip.

After `mamba install argcomplete`, if I run `activate-global-python-argcomplete --user` and restart a new shell (and activate the custom environment), the auto completion does not seem to work. While if I run `eval "$(register-python-argcomplete mintpy)"`, it works as below, but only in the current shell, and the effect is gone when I start a new shell; plus, I don't see the nice script explanation on the right as yours. What did I miss on the activation?

```bash
(insar) yunjunz:~>$ mintpy prep_
prep_aria      prep_cosicorr  prep_fringe    prep_gamma     prep_gmtsar    prep_hyp3      prep_isce      prep_roipac    prep_snap
```

falkamelung:
I quickly tried on the same system and it did not work either for me.

avalentino:
The @yunjunz I think that `activate-global-python-argcomplete --user` puts a script in `.bash_completion.d`, but than that script shall be sourced in the `.bashrc`.
If i remember correctly, some linux distribution does it automatically in the default configuration, while in other cases the `.bashrc` have to be modified to do it.

See, e.g., https://serverfault.com/questions/506612/standard-place-for-user-defined-bash-completion-d-scripts

yunjunz:
Thanks @avalentino, you are right: the `activate-global-python-argcomplete --user` command works for Linux, because the `.bashrc` file sources the `.bash_completion` by default.

For macOS, there is a [bash version compatibility](https://kislyuk.github.io/argcomplete/#global-completion) issue, after installing the latest version of bash, the same activation command works as well!

I still could get the nice script explanation though. Any idea?

PR #1065, opened 2023-08-08
dep/doc: add tab completion via `argcomplete`

yunjunz:

+ add `argcomplete` to the dependency for tab completion capability, which is implemented in #823 and described in #1041. This package is light (only depends on pure Python), and available on pip and conda-forge (as noarch package).
   - requirements.txt: add argcomplete
   - setup.py: move argcomplete from optional to required

+ docs/installation.md: add "tab complete sub-commands" as another sub-section in Post-Installation Setup.

**Reminders**

- [x] Pass [Circle CI test](https://app.circleci.com/pipelines/github/yunjunz/MintPy/2277/workflows/218deb04-7c97-44fd-9add-15044815f908/jobs/2285) (green)

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -230,7 +230,7 @@ export HDF5_USE_FILE_LOCKING=FALSE     # request that HDF5 file locks should NOT
 
 #### d. Tab complete sub-commands ####
 
-If you run mintpy in sub-commands style, e.g. <code>mintpy view</code> instead of <code>view.py</code>, we recommend activating the tab completion: 
+If you run mintpy in sub-commands style, e.g. <code>mintpy view</code> instead of <code>view.py</code>, we recommend activating the tab completion:
 
 ```bash
 activate-global-python-argcomplete --user
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,25 +21,25 @@ mamba install -c conda-forge mintpy
 <details>
 <p><summary>via docker</summary></p>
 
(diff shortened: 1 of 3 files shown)


avalentino (approved):
The changes are good to me.
Probably you may want to point out that, on Mac OSX, `argcomplete >= 3.0` works perfectly on the default shell `zsh`. In that case no workaround is needed.

falkamelung:
I was surprised that you can call mintpy functions using `mintpy` although there is no script/file called `mintpy`. Some explanation of why this works could be helpful.

scottstanie:
> I was surprised that you can call mintpy functions using `mintpy` although there is no script/file called `mintpy`. Some explanation of why this works could be helpful.

These come from the Mintpy [entry points](https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts), which is the typical way a python package can make a command line tool available when you `pip install` the package (without messing with any `$PYTHONPATH` or `$PATH` environment variables). https://setuptools.pypa.io/en/latest/userguide/entry_point.html#console-scripts gives a minimal example to see how that works

yunjunz:
> I was surprised that you can call mintpy functions using `mintpy` although there is no script/file called `mintpy`. Some explanation of why this works could be helpful.

This nice feature was introduced in version 1.5.0 by @avalentino. It's very similar to the latest GMT. I have not advertised it heavily yet, as it may be confusing to users who are used to the `view.py` style.

PR #384, opened 2020-06-22
UAVSAR (stripmapStack) support via YYMMDDTHHMMSS time format

taliboliver:
**Description of proposed changes**

This PR adds UAVSAR support (#376). The detailed changes include:

+ add `YYYYMMDDThhmm` time string support to import to isce processed uavsar coregistered interferogram stack into mintpy.

+ add `get_time_str_format()` for time format identification in utils/ptime.py and objects/stack.py

+ adapt utils.readfile.py and ifgram_inversion.py to read the new time format 

**Reminders**

- [x] Pass Codacy code review (green)
- [x] Pass Circle CI test (green)
- [x] Make sure that your code follows our style. Use the other functions/files as a basis.
- [x] If modifying functionality, describe changes to function behavior and arguments in a comment below the function declaration.
- [x] If adding new functionality, add a detailed description to the documentation and/or an example.

PR #1073, opened 2023-08-18
WIP: Add function to create CF-compliant arrays/metadata for HDF5 stacks

scottstanie:

Start of the implementation described here: https://github.com/insarlab/MintPy/discussions/1016 to make mintpy HDF5 stacks readable by gdal/xarray/qgis.

I'll need to think of a way we can smoothly incorporate this into the `prep_` scripts without interruption.

As a side note: 
>Maybe we could use "datetime" instead? Neither "date" nor "time" feels accurate, given that we handle both spaceborne and airborne data

For now I still have `time` as the possible attribute for the date/datetime stacks just because [that's what the CF-conventions suggested](http://cfconventions.org/cf-conventions/cf-conventions.html#time-coordinate). They also generally have dates/datetime, but use `time` as the standard variable name (even if you have daily granularity). 
Also, it currently has `units=f"days since {str(date_arr[0])}"` as the time units, but this can be `seconds` for intra-day stacks (e.g. for @taliboliver 's deltaX data). But i don't know what modifications he made; i've only seen that there's a `date` dataset usually.

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/src/mintpy/prep_utils.py
+++ b/src/mintpy/prep_utils.py
@@ -1,0 +1,209 @@
+"""Utilities to make HDF5 files CF-compliant and readable by GDAL."""
+from __future__ import annotations
+
+import datetime
+
+import h5py
+import numpy as np
+import pyproj
+
+from mintpy.utils import readfile
+
+__all__ = ["write_coordinate_system"]
+
+
+def write_coordinate_system(
+    filename: str,
+    dset_name: str,
+    xy_dim_names: tuple[str, str] = ("x", "y"),
+    grid_mapping_dset: str = "spatial_ref",
+) -> None:
+    """
+    Write the coordinate system CF metadata to an existing HDF5 file.
+
+    Parameters
+    ----------
+    filename : str
+        File path.
+    dset_name : str
+        Dataset name within the HDF5 file.
+    xy_dim_names : tuple[str, str], optional
+        x and y dimension names, by default ("x", "y").
+    grid_mapping_dset : str, optional
+        Dataset name for grid mapping HDF5 attribute
+        By default "spatial_ref" (matching `rioxarray`).
+    """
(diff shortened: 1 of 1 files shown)


scottstanie:
It's currently failing on one of the integration tests in the `readfile.read_hdf5_file`.
It would be good to add a unit test for this, since the error is saying that `data` is never assigned.
There's only checks for 2D and  higher dimensions..
```
        # 2D dataset
        if ds.ndim == 2:
            ...
            data = ...
```
so it's likely one of the 1D or 0D datasets is tripping it up (but that should be caught separately in `read_hdf5_file` i think, since there shouldn't be a code path to an undefined var)

yunjunz:
Thank you @scottstanie for this exciting PR!!! 

I am trying to review this PR, but it does not seem easy to see what data and metadata have been added in the new format. I have two questions:
1. I assume the [San Francisco Bay dataset](https://github.com/insarlab/MintPy/blob/main/docs/demo_dataset.md#sentinel-1-on-san-francisco-bay-with-aria) from ARIA can be used to test this new capability, right?
2. To document this new format, could you provide a minimal example for creating and examining the old/new format of the HDF5 file? I was relying on `info.py`, but it does not seem to work with the new format yet.

> Also, it currently has units=f"days since {str(date_arr[0])}" as the time units, but this can be seconds for intra-day stacks (e.g. for @taliboliver 's deltaX data). But i don't know what modifications he made; i've only seen that there's a date dataset usually.

For the intra-day stacks, @taliboliver uses the `YYYYMMDDTHHMM` format in the `date` dataset, instead of the usual `YYYYMMDD` format, and [modified the code throughout mintpy](https://github.com/insarlab/MintPy/pull/384) to automatically identify this difference while reading it.

I am not familiar enough with the new changes in this PR yet to help the choice here.

PR #1079, opened 2023-08-27
WIP: changes to `Gps` class to compare to other projections

scottstanie:

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/src/mintpy/objects/gps.py
+++ b/src/mintpy/objects/gps.py
@@ -18,7 +18,13 @@ import numpy as np
 from pyproj import Geod
 
 from mintpy.objects.coord import coordinate
-from mintpy.utils import ptime, readfile, time_func, utils1 as ut
+from mintpy.utils import (
+    ptime,
+    readfile,
+    time_func,
+    utils0 as ut0,
+    utils1 as ut,
+)
 
 UNR_SITE_LIST_FILE = 'http://geodesy.unr.edu/NGLStationPages/DataHoldings.txt'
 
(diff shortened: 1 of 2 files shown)


src/mintpy/objects/gps.py
```diff
@@ -17,8 +17,13 @@
...
 from pyproj import Geod
 
-from mintpy.objects.coord import coordinate
-from mintpy.utils import ptime, readfile, time_func, utils1 as ut
+from mintpy.utils import (
+    ptime,
+    readfile,
+    time_func,
+    utils0 as ut0,
+    utils1 as ut,
```
yunjunz, line 25: `utils0` is imported within `util1` already, so we don't need an extra `utils0 as ut0` here. This is a messy part in mintpy, and the [current logic](https://mintpy.readthedocs.io/en/latest/api/module_hierarchy/) is: 1) `utils1` imports everything from `utils0`; 2) `utils` imports everything from `utils0` and `utils1`.

src/mintpy/objects/gps.py
```diff
@@ -508,10 +509,30 @@ def get_los_geometry(self, geom_obj, print_msg=False):
             az_angle  = ut.heading2azimuth_angle(float(geom_obj['HEADING']))
 
         else:
-            raise ValueError(f'input geom_obj is neight str nor dict: {geom_obj}')
+            raise ValueError(f'input geom_obj is neither str nor dict: {geom_obj}')
 
-        return inc_angle, az_angle
+        return inc_angle or np.nan, az_angle or np.nan
 
+    def get_image_values(self, filename, datasetName=None, pad: int = 0, print_msg=False):
```
yunjunz, line 516: This function is not used currently anywhere, so it's fine with me. Just to be clear: from our experience, given a GNSS site, we either interpolate InSAR or average multiple InSAR pixels around the GNSS site. Personally, I have not gotten a good result yet by comparing GNSS and its nearest InSAR pixel yet.

src/mintpy/objects/gps.py
```diff
@@ -508,10 +509,30 @@ def get_los_geometry(self, geom_obj, print_msg=False):
             az_angle  = ut.heading2azimuth_angle(float(geom_obj['HEADING']))
 
         else:
-            raise ValueError(f'input geom_obj is neight str nor dict: {geom_obj}')
+            raise ValueError(f'input geom_obj is neither str nor dict: {geom_obj}')
 
-        return inc_angle, az_angle
+        return inc_angle or np.nan, az_angle or np.nan
```
yunjunz, line 514: I am just curious: in this syntax, when will the `np.nan` value be returned? Is it the same scenario as the `raise ValueError` above?

yunjunz (commented):
Thank you @scottstanie for the PR! My main comment is the `get_image_rowcol()`: since it relies on `mintpy.objects.coord.coordinate.geo2radar()`, and [`mintpy.objects.coord.py` relies on `mintpy.utils.utils0.py`](https://github.com/insarlab/MintPy/blob/7ff149b7a99ece61f0fee33482f969a6540ba279/src/mintpy/objects/coord.py#L16), I would suggest moving this function into `mintpy.objects.coord.py` to avoid the circular imports. You could find the complete reliance info among sub-modules in mintpy [here](https://mintpy.readthedocs.io/en/latest/api/module_hierarchy/). The possible location could be:

+ as an independent function as it's now OR
+ inside `geo2radar()` OR
+ inside `lalo2yx()`

UPDATE: If we assume the InSAR file is already geocoded (in EPSG:4236 or other projections), then merging `geo_iamge_rowcol()` into `lalo2yx()` sounds the simplest to me. What do you think?

Could you also rebase this PR against the latest main branch?

PR #1084, opened 2023-09-07
view: allow unit change when viewing seasonal time functions

bbuzz31:
small change to allow a user to change the units when viewing both annualAmplitude[Phase] and semiAnnualAmplitude[Phase]

e.g., `view.py geo_velocity.h5 annualAmplitude semiAnnualAmplitude -u mm`


yunjunz (approved):
LGTM. Thank you @bbuzz31.

PR #1129, opened 2023-12-04
replace `setup.py` with `pyproject.toml`

yunjunz:

+ add `tests/requirements.txt` file for test requirements

+ `requirements.txt`: move pre-commit into the `tests/requirements.txt` file

+ `pyproject.toml`: add all the corresponding contents from `setup.py` based on https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

+ delete `setup.py`

**Reminders**

- [x] Pass [Circle CI test](https://app.circleci.com/pipelines/github/yunjunz/MintPy/2432/workflows/56e6adbb-9cf6-4317-abc0-894a698806ad/jobs/2438) (green)


scottstanie:
I think the `python>=3.6` part isn't install-able via pip, so it had to be removed from `requirements.txt`. It's enforced by having it in `pyproject.toml`.
(sorry I didn't mean to push directly to you branch! 😬 )

By the way, had you considered bumping up the min python version? Both 3.6 and 3.7 are now in the [end-of-life](https://devguide.python.org/versions/#versions) stage

requirements.txt
```diff
@@ -1,20 +1,15 @@
-# requirements.txt          for dependency installation via conda
-# docs/ports.txt            for dependency installation via mac-ports
-# setup.py                  for mintpy     installation via pip after the dependency installation above
 python>=3.6
```
avalentino: I think that this is the problem.
This is OK for `conda` but not for `pip`.
I understand that this file is also include via `pyptoject.toml` dynamic stuff but `pip` cannot resolve a package named `python`.

My recommendation is to remove `python` from the requirement file and maybe add directly to the CI configuration file creating the `conda` environment.

yunjunz: That was the issue as @scottstanie has pointed out. I see that [circle CI has successfully installed python 3.10](https://app.circleci.com/pipelines/github/yunjunz/MintPy/2433/workflows/5726d71d-9396-45a7-a0a6-ce317e1e5dfa/jobs/2439?invite=true#step-103-15059_16) as well. So I guess we could remove this completely here.

yunjunz:
Thank you @scottstanie for the quick fix. 

I just bumped the min version to 3.8 as you suggested. I will look at the new syntax of 3.7 and 3.8 on another day to see what can be simplified in the code to leverage the new version!

PR #1146, opened 2024-02-15
`view`: support scatter plot via `--style scatter` option

falkamelung:

This PR supports the scatter plot via `matplotlib.pyplot.scatter()`, in addition to the existing image plot via `matplotlib.pyplot.imshow()`, for `view.py` single plot scenario. This is useful for high-resolution data with lots of holes from unsuccessful unwrapping. Plotting larger than pixel size fills holes like a low-pass filter. 

This is controlled via two options: `--style` with default of image, and `--scatter-size` with default of 10 point (**2), respectively.

In the image, top is  regular view.py (using imshow) and bottom is `--style scatter --scatter-size 8`
image

**Reminders**

### DIFF
--- a/src/mintpy/cli/view.py
+++ b/src/mintpy/cli/view.py
@@ -78,10 +78,13 @@ def create_parser(subparsers=None):
                              '  sqrt    = x^1/2\n'
                              '  reverse = x * -1\n'
                              '  inverse = 1 / x')
-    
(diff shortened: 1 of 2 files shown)


yunjunz (approved):
Thank you @falkamelung for the PR. The new style looks nice!

PR #1158, opened 2024-03-11
`prep_nisar`: update for changes in NISAR GUNW products

mirzaees:

- The mask file is updated in GUNW products which is corrected
- There was a bug related to projection that subset was not reading bounding box in the correct order. 
- A fix is introduced to correctly add projection to the metadata

### DIFF
--- a/src/mintpy/prep_nisar.py
+++ b/src/mintpy/prep_nisar.py
@@ -266,16 +266,16 @@ def read_and_interpolate_geometry(gunw_file, dem_file, xybbox, mask_file=None):
 
     subset_rows = len(ycoord)
     subset_cols = len(xcoord)
-    
+
     Y_2d, X_2d = np.meshgrid(ycoord, xcoord, indexing='ij')
     bounds = (min(xcoord), min(ycoord), max(xcoord), max(ycoord))
     output_projection = f"EPSG:{dst_epsg}"
-    
-    # Warp DEM to the interferograms grid 
+
+    # Warp DEM to the interferograms grid
     input_projection = f"EPSG:{dem_src_epsg}"
     output_dem = os.path.join(os.path.dirname(dem_file), 'dem_transformed.tif' )
     gdal.Warp(output_dem, dem_file, outputBounds=bounds, format='GTiff',
-              srcSRS=input_projection, dstSRS=output_projection, resampleAlg=gdal.GRA_Bilinear, 
+              srcSRS=input_projection, dstSRS=output_projection, resampleAlg=gdal.GRA_Bilinear,
               width=subset_cols, height=subset_rows,
               options=['COMPRESS=DEFLATE'])
 
@@ -295,11 +295,11 @@ def read_and_interpolate_geometry(gunw_file, dem_file, xybbox, mask_file=None):
             mask_src_epsg = int(proj.GetAttrValue('AUTHORITY', 1))
             del mask_dataset
 
-            # Warp mask to the interferograms grid 
+            # Warp mask to the interferograms grid
             input_projection = f"EPSG:{mask_src_epsg}"
             output_mask = os.path.join(os.path.dirname(mask_file), 'mask_transformed.tif' )
             gdal.Warp(output_mask, mask_file, outputBounds=bounds, format='GTiff',
-              srcSRS=input_projection, dstSRS=output_projection, resampleAlg=gdal.GRA_Byte, 
+              srcSRS=input_projection, dstSRS=output_projection, resampleAlg=gdal.GRA_Byte,
               width=subset_cols, height=subset_rows,
               options=['COMPRESS=DEFLATE'])
 
@@ -442,9 +442,9 @@ def prepare_stack(
 
             # read/write perpendicular baseline file
             f['bperp'][i] = dataset['pbase']
-                
+
             prog_bar.update(i + 1, suffix=date12_list[i])
         prog_bar.close()
-    
+
     print(f"finished writing to HDF5 file: {outfile}")
     return outfile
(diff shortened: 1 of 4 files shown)


src/mintpy/prep_nisar.py
```diff
@@ -195,8 +204,8 @@ def common_raster_bound(input_files, utm_bbox=None):
         x_bounds.append([west, east])
         y_bounds.append([south, north])
     if not utm_bbox is None:
-        x_bounds.append([utm_bbox[0], utm_bbox[2]])
-        y_bounds.append([utm_bbox[1], utm_bbox[3]])
+        x_bounds.append([utm_bbox[1], utm_bbox[3]])
+        y_bounds.append([utm_bbox[0], utm_bbox[2]])
```
yunjunz, line 208: I would suggest to modify `bbox_to_utm()` output to be in the order of (x0, y0, x1, y1) instead of modifying here, to be consistent with other bbox instances in the code.

yunjunz (commented):
Thanks @mirzaees. Please check the two minor comments below. @scottstanie could you take a look at the rest of the change?

src/mintpy/prep_nisar.py
```diff
@@ -61,9 +61,9 @@ def load_nisar(inps):
     print(f"Found {len(input_files)} unwrapped files")
 
     if inps.subset_lat:
-        bbox = (inps.subset_lat[0], inps.subset_lon[0], inps.subset_lat[1], inps.subset_lon[1])
+        bbox = (inps.subset_lon[0], inps.subset_lat[0], inps.subset_lon[1], inps.subset_lat[1])
```
scottstanie, line 64: so this isn't converting from SNWE to (left, bottom, right top), it's just fixing the ordering for (left, bottom, right, top)?

scottstanie (approved):

src/mintpy/prep_nisar.py
mirzaees: yes the input arguments follows mintpy subsetting with two arguments `--sub-lat` and `--sub-lon `. They are ordered to make bbox here

hfattahi (commented):
Thanks @mirzaees . Please see my comments below.

src/mintpy/prep_nisar.py
```diff
@@ -263,20 +274,47 @@ def read_and_interpolate_geometry(gunw_file, dem_file, xybbox):
     # dem_subset_array = np.zeros((subset_rows, subset_cols), dtype=raster_array.dtype)
     Y_2d, X_2d = np.meshgrid(ycoord, xcoord, indexing='ij')
 
-    if not src_epsg == dst_epsg:
-        coord_transform = Transformer.from_crs(dst_epsg, src_epsg, always_xy=True)
+    if not dem_src_epsg == dst_epsg:
+        coord_transform = Transformer.from_crs(dst_epsg, dem_src_epsg, always_xy=True)
```
hfattahi, line 278: This comment did not make it to my review so posting it here again. It looks like you want to check if the epsg of the input DEM is different from GUNW and if they are different then project the DEM to the GUNW epsg and to its exact grid or a given bounding box. I'm wondering why you are not letting gdal.warp do that for you. Something like this should work:
```
gdal.Warp(output_dem_path, 
          src_ds, 
          format='GTiff', 
          outputBounds=bbox, 
          xRes=x_res, yRes=y_res,
          srcSRS=source_epsg, 
          dstSRS=target_epsg,
          resampleAlg=gdal.GRA_Bilinear,
          options=['COMPRESS=DEFLATE']) 
```

src/mintpy/prep_nisar.py
```diff
@@ -263,20 +274,47 @@ def read_and_interpolate_geometry(gunw_file, dem_file, xybbox):
...
+        try:
+            mask_dataset = gdal.Open(mask_file, gdal.GA_ReadOnly)
+            geotransform = mask_dataset.GetGeoTransform()
+            proj = gdal.osr.SpatialReference(wkt=mask_dataset.GetProjection())
+            mask_src_epsg = int(proj.GetAttrValue('AUTHORITY', 1))
+            mask_raster_array = mask_dataset.ReadAsArray()
+            del mask_dataset
+
+            if not mask_src_epsg == dst_epsg:
+                coord_transform = Transformer.from_crs(dst_epsg, mask_src_epsg, always_xy=True)
```
hfattahi, line 304: Similar to my comment above on DEM transformation , I think you could use gdal.warp here as well. Since your mask is binary, you may use nearest neighbor interpolation.

src/mintpy/prep_nisar.py
```diff
@@ -61,16 +61,16 @@ def load_nisar(inps):
...
+        bbox = None
 
     # extract metadata
     metadata, bounds = extract_metadata(input_files, bbox)
 
     # output filename
-    stack_file = os.path.join(inps.out_dir, "inputs/ifgramStack.h5")
-    geometry_file = os.path.join(inps.out_dir, "inputs/geometryGeo.h5")
+    stack_file = os.path.join("./inputs/ifgramStack.h5")
+    geometry_file = os.path.join("./inputs/geometryGeo.h5")
```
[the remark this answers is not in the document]
mirzaees: The reason I removed it was a confusing path set in load_data where it pointed to '../mintpy' and it would show an error if you select another folder name. But you are right, I could keep this general format and fix it in load_data

src/mintpy/prep_nisar.py
mirzaees: Yes thank you, that is a more proper way. I will change it

mirzaees (commented):
Thanks @hfattahi @scottstanie and @yunjunz 
I changed as you suggested, let's wait for some good data to test before merging

yunjunz (approved):

hfattahi (approved):
I agree that it will be great to test with updated GUNW before merging this PR. @jplzhan and @katia-tymofyeyeva are working on updating the GUNW test data.

PR #1159, opened 2024-03-13
utils0: add `epsg_code2utm_zone()`

yunjunz:

+ utils.utils0.py: add `epsg_code2utm_zone()` to convert the EPSG code to the UTM zone string.

**Reminders**

### DIFF
--- a/src/mintpy/utils/utils0.py
+++ b/src/mintpy/utils/utils0.py
@@ -295,7 +295,7 @@ def utm_zone2epsg_code(utm_zone):
 
 
 def epsg_code2utm_zone(epsg_code):
-    """Convert EPSG code (in UTM coordinates) to UTM Zone string.
+    """Convert EPSG code to UTM Zone string.
 
     Parameters: epsg_code - str / int, EPSG code
     Returns:    utm_zone  - str, atr['UTM_ZONE'], comprises a zone number
--- a/src/mintpy/utils/utils0.py
+++ b/src/mintpy/utils/utils0.py
@@ -282,7 +282,47 @@ def utm_zone2epsg_code(utm_zone):
         'south': utm_zone[-1].upper() == 'S',
     })
     epsg_code = crs.to_authority()[1]
+
+    # Link: https://gis.stackexchange.com/questions/365584/convert-utm-zone-into-epsg-code
+    '''
+    zone = '36'
+    south = True
+    epsg_code = 32600
+    epsg_code += int(zone)
+    if south is True:
+        epsg_code += 100
+    '''
+
     return epsg_code
+
+
+def epsg_code2utm_zone(epsg_code):
+    """Convert EPSG code (in UTM coordinates) to UTM Zone string.
+
+    Parameters: epsg_code - str / int, EPSG code
+    Returns:    utm_zone  - str, atr['UTM_ZONE'], comprises a zone number
+                            and a hemisphere, e.g. 11N, 36S, etc. None for
+                            a EPSG code not in a UTM coordnate system
+    Examples:   utm_zone = epsg_code2utm_zone('32736')
+    """
+    from pyproj import CRS
+    crs = CRS.from_epsg(epsg_code)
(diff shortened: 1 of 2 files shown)


src/mintpy/utils/utils0.py
```diff
@@ -282,9 +282,45 @@ def utm_zone2epsg_code(utm_zone):
...
+    """Convert EPSG code (in UTM coordinates) to UTM Zone string.
+
+    Parameters: epsg_code - str / int, EPSG code
+    Returns:    utm_zone  - str, atr['UTM_ZONE'], comprises a zone number
+                            and a hemisphere, e.g. 11N, 36S, etc. None for
+                            a EPSG code not in a UTM coordnate system
+    Examples:   utm_zone = epsg_code2utm_zone('32736')
+    """
+    from pyproj import CRS
+    crs = CRS.from_epsg(epsg_code)
```
scottstanie, line 307: If you wanted, you could even make this `CRS.from_user_input` to accept `"EPSG:32617"` or `32617` or `"32617"` 

```

In [15]: CRS.from_epsg(32617) == CRS.from_user_input(32617) == CRS.from_user_input("32617") == CRS.from_user_input("epsg:32617")
Out[15]: True
```
and it doesn't seem any slower


```
In [8]: %timeit CRS.from_epsg(32617).utm_zone
472 µs ± 3.58 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [9]: %timeit CRS.from_user_input('EPSG:32617').utm_zone
475 µs ± 4.06 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [10]: %timeit CRS.from_user_input('32617').utm_zone
478 µs ± 9.87 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

src/mintpy/utils/utils0.py
```diff
@@ -282,9 +282,45 @@ def utm_zone2epsg_code(utm_zone):
...
+        print(f'WARNING: input EPSG code ({epsg_code}) is NOT a UTM zone, return None and continue.')
+
+    # epsg_code = str(epsg_code)
+    # if epsg_code.startswith('326'):
+    #     utm_zone = epsg_code[3:] + 'N'
+    # elif epsg_code.startswith('327'):
+    #     utm_zone = epsg_code[3:] + 'S'
+    # else:
+    #     utm_zone = None
+    #     print(f'WARNING: input EPSG code ({epsg_code}) is NOT a UTM zone, return None and continue.')
```
scottstanie, line 319:
```suggestion
```
delete?

scottstanie (approved):

src/mintpy/utils/utils0.py
yunjunz: I am not sure about how the conversion between ESPG code and UTM Zone is done mathematically, then saw the code from [Sara's PR](https://github.com/insarlab/MintPy/pull/1158/files#diff-86e10c5886296273b3776c7f4e642ef2b4abff0929994b2ace1e5817b0faffacR145-R148), and thought this is helpful for understanding. Is it correct?

src/mintpy/utils/utils0.py
yunjunz: That's good to know. I added them in the comment. Thanks.

src/mintpy/utils/utils0.py
mirzaees: This might help:
https://docs.up42.com/data/reference/utm#:~:text=For%20the%20Northern%20Hemisphere,find%20the%20UTM%20zone%20first.

yunjunz: Thank you @mirzaees, this is very helpful! I added the link to the inline comment. Now the EPSG code feels more generic and simple. It looks like we could use EPSG code for all kinds of calculations and checking very easily, without UTM_ZONE at all. Maybe we could remove this metadata completely, what do you think? @mirzaees  @scottstanie @hfattahi

PR #1180, opened 2024-04-16
Add HDF-EOS5 metadata for HyP3 products

### LINKED ISSUE #1162: `hyp3`: more metadata for `HDF-EOS5` support
**Description of the desired feature**

[`HDF-EOS5` file format](https://mintpy.readthedocs.io/en/latest/hdfeos5/) is an easy way to save the final time-series product into a single file. There is now near-automatic support for that after https://github.com/insarlab/MintPy/pull/1161. However, the following metadata is missing. It will be great to provide them from the ASF HyP3 side:
   - relativeOrbit: required by HDF-EOS5
   - startUTC: used to calculate the first frame number
   - stopUTC: used to calculate the last frame number
   - ascendingNoteTime: use to calculate the first/last frame number [updated]

Then, all hyp3 products can be saved in HDF-EOS5 format, if the user just turn on the `mintpy.save.hdfEos5` option, without any manual adding of metadata (as I did in [here](https://github.com/yunjunz/MintPy/blob/51fe70e6e85fc62620160730be96ceff82e32053/tests/configs/RidgecrestSenDT71.txt#L26)).

Let me know what you think please. @forrestfwilliams @talogan @jhkennedy 

Next, we would need to update the [example ASF HyP3 dataset](https://github.com/insarlab/MintPy/blob/main/docs/demo_dataset.md#sentinel-1-of-the-2019-ridgecrest-california-earthquake-sequence-with-hyp3) (to include the above metadata); and cut for a new release of mintpy.

forrestfwilliams:
At @yunjunz's suggestion, this PR adds the `startUTC`, `stopUTC`, and `relativeOrbit` fields to HyP3-derived products to support HDF-EOS5 conversion.

**Reminders**

- [x] Fix #1162


forrestfwilliams:
@yunjunz not sure I got `startUTC` and `stopUTC` right here. Should these be attributes of each interferogram, or are they the start/stop time for the entire stack (i.e., min input SLC date, and max input SLC date)?

src/mintpy/prep_hyp3.py
```diff
@@ -109,9 +109,13 @@ def add_hyp3_metadata(fname, meta, is_ifg=True):
     # HDF-EOS5 metadata
     if hyp3_meta['ReferenceGranule'].startswith('S1'):
         meta['beam_mode'] = 'IW'
+
+    meta['startUTC'] = date1.strftime('%Y-%m-%d %H:%M:%S.%f') + '+00:00'
+    meta['stopUTC'] = date2.strftime('%Y-%m-%d %H:%M:%S.%f') + '+00:00'
```
yunjunz, line 114: [Here](https://github.com/insarlab/MintPy/blob/32645dd5df20ae30583ae0e4d528bcb71fd5fef6/src/mintpy/utils/isce_utils.py#L203-L205) is how we calculate the ASF frame number for Sentinel-1. So it seems like we need the ascending node time as well.

As for the current start/stopUTC, why add the `+00:00` at the end? Based on the example below, I would think the `*.strftime()` is enough.

```bash
(insar) yunjunz:~/data/test/FernandinaSenDT128/mintpy>$ info.py velocity.h5 
...
  startUTC                                        2014-12-13 11:49:44.367659
  stopUTC                                         2014-12-13 11:50:07.299273
...
```

src/mintpy/prep_hyp3.py
```diff
@@ -109,9 +109,13 @@ def add_hyp3_metadata(fname, meta, is_ifg=True):
...
+    meta['relativeOrbit'] = ((int(hyp3_meta['ReferenceOrbitNumber']) - 73) % 175) + 1
```
yunjunz, line 115: The relative orbit calculation is different for S1A and S1B, as shown in [`s1_reader`](https://github.com/isce-framework/s1-reader/blob/bb585c77ee538324212085f4737f3d2d534ac998/src/s1reader/s1_burst_slc.py#L1115-L1118).

yunjunz (changes requested):
Hi @forrestfwilliams, start/stopUTC should come from the same acquisition, so the current calculation should be fine. 

However, we would also need the ascending note time of the same acquisition to calculate the ASF frame number for S1. I guess that could be added later in the HyP3 metadata? Regardless, we could merge this PR without this extra attribute.

src/mintpy/prep_hyp3.py
forrestfwilliams: The `+00:00` just indicates that this timestamp is in the UTC timezone. If don't need that though, happy to remove it.

src/mintpy/prep_hyp3.py
forrestfwilliams: On it!

src/mintpy/prep_hyp3.py
yunjunz: I see. Yes, let's remove it for consistency please.

src/mintpy/prep_hyp3.py
forrestfwilliams: Looks like the example HyP3 data is an older version that doesn't include some metadata fields. I'll create a new version.

yunjunz:
> Hi @forrestfwilliams, start/stopUTC should come from the same acquisition, so the current calculation should be fine.

My misunderstanding: the `start/stopUTC` should come from the same acquisition, i.e. the starting and ending time of one image. So the current calculation is NOT correct. We could grab the two time stamps from the Reference or Secondary file name directly.

yunjunz (approved):
Looks all good to me now. Thank you @forrestfwilliams.

Update: beam_swath is needed for hyp3-gamma/isce2 products.

forrestfwilliams:
Sounds good! Do we need to update `startUTC`/`stopUTC`, or are they correct now?

PR #1195, opened 2024-05-12
parallel downloads

ritwika21:

**Reminders**

- [ ] Fix #xxxx

### DIFF
--- a/src/mintpy/tropo_pyaps3.py
+++ b/src/mintpy/tropo_pyaps3.py
@@ -466,7 +466,21 @@ def check_pyaps_account_config(tropo_model):
 
 
 ###############################################################
+from concurrent.futures import ThreadPoolExecutor
+
 def dload_grib_files(grib_files, tropo_model='ERA5', snwe=None):
+    # import pdb; pdb.set_trace()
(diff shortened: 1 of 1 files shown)


yunjunz:
Thank you @ritwika21 for contributing! 

Could you add some description to the PR?

And some questions on the ERA5 parallel downloading: 
1. How much time does it save now with this PR, compared with the current version, with an example?
2. Based on what I was aware of a couple of years ago, ECMWF (via the Copernicus Climate Data Store) allows a max of 3 submitted jobs per user at the same time. Will it make more sense to set the parallel job number to 3, instead of 64, if the 3-job-limit still exists?

yunjunz:
pre-commit.ci autofix

falkamelung:
I heard that downloading one date as one file is not the right approach.  ECMWF might have options to download all the required days for one SAR dataset as one file.   I asked some atmospheric scientists for help, and they were shocked about how we download the data.  But I did not pursue this yet.

yunjunz:
Any link to the code or documentation for the downloading all-at-once approach will be very helpful.

PR #1210, opened 2024-05-31
`save_kmz`: support data in UTM

### LINKED ISSUE #1211: Output KMZs not properly geocoded for time series in UTM
**Description of the problem**

KMZs produced when running the `google_earth` step are not properly geocoded if data are in UTM because `save_kmz.py` assumes that data are in WGS84 and creates an invalid bounding box.

https://github.com/insarlab/MintPy/pull/1210

**Full script that generated the error**

```
smallbaselineApp.py path/to/config --work-dir path/to/work_dir --dostep google_earth
```

**Error example**
The KMZ should appear in Google Earth like this:
image

But, if data are in UTM, it appears like this:

image

Adjusting Google Earth preferences to use UTM does not work because the kmz is not correctly geocoded.

**System information**

* Operating system: Linux 
* Python environment: conda 
* MintPy version: 1.6.0 
* InSAR processor/product: hyp3 
* Your custom / default template file (if the bug is related to a specific dataset):

Alex-Lewandowski:

This addresses an issue causing improperly geocoded velocity kmz files from data in UTM. KML only natively supports WGS84, so I added code to convert the bounding box coordinates to WGS84 when necessary. 

**Reminders**

- [x] Fix #1211 

### DIFF
--- a/src/mintpy/save_kmz.py
+++ b/src/mintpy/save_kmz.py
@@ -13,43 +13,61 @@ from zipfile import ZipFile
 
 import numpy as np
 from lxml import etree
-from matplotlib import colorbar, colors, pyplot as plt, ticker
+from matplotlib import colorbar, colors
+from matplotlib import pyplot as plt
+from matplotlib import ticker
 from pykml.factory import KML_ElementMaker as KML
 
 import mintpy
 from mintpy import subset
-from mintpy.utils import (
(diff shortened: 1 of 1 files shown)


yunjunz (approved):
Looks good to me. Thank you @Alex-Lewandowski!

PR #1216, opened 2024-06-11
`readfile`: support `Gamma` SLC in complex32 format

yunjunz:

+ utils.readfile: support Gamma slc in complex32 format
   - add scomplex to complex32 data type conversion
   - read_binary_file(): support .rslc/rmli file extension
   - read_complex_int16(): refactor to use cpx_band in str for a more flexible output data; revise the output from 2 matrices into 1 matrix
   - read_binary(): support complex32 while calling read_complex_int16(), since it's not supported in numpy.

+ utils.readfile.auto_no_data_value(): add geometrical offset files from alos2App with num of band = 1

+ view.viewer.plot(): set pixels with no data value to np.nan, instead of call np.ma.masked_where(), which seems over-complicated (compared with setting to nan) and I could not recall the reason anymore.

+ cli/prep_isce.py: set -f default from str to list, to be consistent with the non-default var type.

**Reminders**

- [ ] Pass [Circle CI](https://app.circleci.com/jobs/github/yunjunz/MintPy/2649) test (green)
- [x] Pass local test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,30 +53,35 @@ jobs:
           command: |
             mkdir -p ${HOME}/data
             ${MINTPY_HOME}/tests/smallbaselineApp.py --dir ${HOME}/data --dset FernandinaSenDT128
+            rm -r ${HOME}/data
 
       - run:
           name: Integration Test 2/4 - SanFranSenDT42 (ARIA)
           command: |
             mkdir -p ${HOME}/data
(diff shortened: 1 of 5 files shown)


yunjunz:
Local testing passed without issue. The failed [circle CI test](https://app.circleci.com/pipelines/github/yunjunz/MintPy/2646/workflows/4769e1dd-2405-4f11-8b18-2161da93f40c/jobs/2653) shows that GDAL can not be imported, but it's installed during the installation process. Nevertheless, the circle CI error is not related with the PR, and probably due to some issue on circle CI itself.

PR #1238, opened 2024-07-25
`subset.py`: support `--lat/lon` in radar coordinates

### LINKED ISSUE #1232: Support of --lat --lon for subsetting files in radar coordinates
**Description of the desired feature**

Subset.py did not support extracting from a file in radar coordinates a subset by giving lat/lon coordinates. This PR addresses this. This is useful to extract data for individual buildings after locating them in Google Earth or similar.

The accuracy of this approach is unclear.  Because of this there is default  buffer used which in the case I looked at was 20 pixels ( 'buf = 2 * np.max(np.abs([x_res, y_res]))' ). For this particular subsetting (radar coordinates, -lat -lon) this buffering is now switched off.

I previously observed that subsets don't always agree with the area you are asking which may be related to this buffer. I will observe and report on this later. If the buffer turns out to be a problem we may want to consider reducing or removing it.

This example demonstrates the improvement from buf=0 in the suggested code.

image

**Are you willing to help implement and maintain this feature?**

- [X] Yes
- [ ] No

falkamelung:
Description of proposed changes
This PR addresses this issue:

https://github.com/insarlab/MintPy/issues/1232

**Reminders**

- [x] Fix #1232


ditafaith:
> Thank you @falkamelung for the PR. Your test on the buffer at #1232 is also very helpful. I modified your code a little bit to make them more compact. The changes look all good to me now.

It seems good. Do I need to update the current version?

PR #1292, opened 2024-11-12
Removes np.nan in output of spatial_filter.py

bjmarfito:

Removes np.nan when using spatial_filter.py. This is useful to generate velocity using timeseries2velocity.py.

**Reminders**

- [ ] Fix #xxxx

## Summary by Sourcery

Bug Fixes:
- Remove np.nan values from the output of spatial_filter.py to ensure compatibility with timeseries2velocity.py.

### DIFF
--- a/src/mintpy/spatial_filter.py
+++ b/src/mintpy/spatial_filter.py
@@ -195,6 +195,7 @@ def filter_file(fname, ds_names=None, filter_type='lowpass_gaussian', filter_par
             data = filter_data(data, filter_type, filter_par)
 
         # save
+        data *= ~np.isnan(data)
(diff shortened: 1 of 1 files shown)


yunjunz (changes requested):
I don't get the intention here, this line of change won't do anything, from my simple testing. Could you clarify? and add some inline comment or a new option instead?

bjmarfito:
Hi @yunjunz, after looking at the code, I found out that adding this line of code does not solve the issue of generating velocity using timeseries2velocity.py after using spatial_filter.py. Need to look  in-depth regarding this problem.

Closing this PR.

PR #1328, opened 2025-03-05
Load ARIA-tools generated ionosphere stack from the template file

ehavazli:
- Add the `--iono` option to `load_data.py` arguments so when ionosphere file is given in the template, the `load_data` step will read the `ionoStack.vrt`.
- Fix a bug in `prep_aria.py` which was causing the script to skip loading to `ionStack.h5`

### DIFF
--- a/src/mintpy/load_data.py
+++ b/src/mintpy/load_data.py
@@ -740,6 +740,7 @@ def prepare_metadata(iDict):
             '--incidence-angle'     : 'mintpy.load.incAngleFile',
             '--azimuth-angle'       : 'mintpy.load.azAngleFile',
             '--water-mask'          : 'mintpy.load.waterMaskFile',
+            '--iono'                : 'mintpy.load.ionUnwFile',
         }
 
         for arg_name, opt_name in ARG2OPT_DICT.items():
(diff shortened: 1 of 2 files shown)


ehavazli:
Here is additional context for the bug fix in `prep_aria.py`:

The issue arises when `smallbaselineApp.py` is run directly with a template that includes `ionoStack.vrt` from an ARIA GUNW stack. By default, `smallbaselineApp.py` calls `prep_aria.py` with the `--update` flag, which causes it to skip loading certain files if their output already exists. Since `ionStack.h5` is formatted the same as `ifgramStack.h5`, the `prep_aria.py` call assumes the output file name as `ifgramStack.h5` and skips loading.

However, since `ionStack.h5` is never created, the workflow fails at the `correct_ionosphere` step due to the missing file with `TypeError: sequence item 7: expected str instance, NoneType found`. This fix reorders the logic to define `out_name` before the run_or_skip() call, ensuring the conditional correctly considers `ionStack.h5` and triggers its loading if needed.

@yunjunz I don't see a specific ARIA test under the `tests` folder. Please let me know if I should add a `tests/test_load_data.py`

PR #1338, opened 2025-04-01
add `mintpy.load.compression = default` option & allow custom compression for geometry HDF5 file

Merdele:

changed 'lzf' to iDict['compression']

loading of geometry radar.h5 was choosing 'lzf' as the compression method despite specifying something else in the config file. 

**Reminders**

## Summary by Sourcery

Bug Fixes:
- Fixed hardcoded compression method when saving geometry radar data to .h5 files, now using the compression method specified in the configuration dictionary

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
         exclude: tests/data/
 
   - repo: https://github.com/PyCQA/isort
-    rev: "6.0.0"
+    rev: "6.0.1"
     hooks:
       - id: isort
         name: sort imports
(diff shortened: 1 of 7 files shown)


yunjunz:
This `lzf` compression is hardwired for geometry files on purpose, for a significantly smaller file size, without much impact on the performance. Did this cause a problem?

Merdele:
Did not cause a problem, needed to the h5 file to not be compressed for other purposes and specifying None in the CLI didn't change it so i thought it was a bug!

yunjunz:
Sorry for the late feedback, I got busy at work. 

I think the following behavior is what we can both be happy with: by default (if the user did not manually specify the compression from the configuration file), use `None` for the stack and `lzf` for the geometry; otherwise, follow and compress as specified.

yunjunz (approved):

PR #1486, opened 2026-03-31
support OPERA level-4 ZTD products for tropospheric correctioin

dbekaert:
This is an update to the PR that was open before: #1473.

Additionally:
- in sync back with main branch
- changed the support to only download from ASF what is needed to cover the user bbox in vertical and lateral space.
This brings down resources drastically in terms of disk need and time to download. Example here is for MExico city example for 6 NISAR acquisitions.
```
after:  du -sh OPERA: 260K 
before:  du -sh OPERA: 17G
```
Comparing the old full 3D method with the new 2D/1D combo (changed to simpler interpolation) shows also nearly identical results.
```
Max abs diff: 0.000853 m
Mean abs diff: 0.000049 m
RMS diff: 0.000090 m
```

- [x] Fix #1473 

## Summary by Sourcery

Add support for tropospheric delay correction using OPERA zenith delay products and wire it into the MintPy workflow and CLI.

New Features:
- Introduce an OPERA-based tropospheric correction workflow that downloads, subsets, and applies OPERA zenith delay products to generate delay time series and corrected displacement files.
- Expose a new tropo_opera CLI entry point and parser, and register it in the main MintPy command-line interface and workflow.
- Add configuration options for selecting the OPERA tropospheric delay method and specifying the OPERA data directory in smallbaselineApp templates.

Enhancements:
- Limit ASF OPERA downloads to the spatial and vertical extent required by the user's geometry and DEM to reduce data volume.


dbekaert:
@yunjunz this should bring the PR back in sync with dev, and also it addresses all the concerns of the bots. 
Note the one codacy issue is the same as how its currently done in mintpy for other functions.
To add subsetting i added the new dependencies for fsspec. Though a fallback is to download the complete file using ASF search.

dbekaert:
Thanks for checking @sssangha. issues for the test are addressed as well.

dbekaert:
@sssangha indeed a bit of a circular codacy issue... @yunjunz we should be good to go again. @sssangha confirmed support for different projections now.

hfattahi:
Thanks @dbekaert and @sssangha for this PR. Overall looks good to me. I had a small stack of NISAR GUNW data in mintpy and decided to compare the tropospheric delay in NISAR GUNW with those computed from opera tropo products using this PR. The two tropospheric delay agree to some great extent. Looking at the difference between the two tropospheric delay, I see very small local differences of 2-3 mm in some regions. I know the difference is within the uncertainty of these model based tropospheric delay estimates, however I believe we are using the same underlying software and same atmospheric model for both NISAR GUNW and OPERA ZTD. I'm not sure what may cause the difference! Maybe direct estimation of slant delay in GUNW while computing ZTD in opera and then mapping to slant ?! I was wondering if you have done similar analysis and if you have any thoughts on this level of differences?

Here is the timeseries of tropo delay from NISAR GUNW as derived from [this PR](https://github.com/insarlab/MintPy/pull/1487) 

tropo_nisar

And here is the timeseries of the opera tropo delay:

tropo_opera

And here is the difference of the two:

nisar_diff_opera_tropo

dbekaert:
hi @hfattahi, thats great. I had not compared it yet with the tropo within the GUNW, though its on the list. This looks good and indeed this is within the model uncertainties.  I suspect this is pretty much related to interpolation differences. A past study from ECMWF for ESA (will see if i can dig up this report) showed that interpolation introduced like to 10% delay variations when they were comparing delays computed starting from pressure and model levels of the same model. 

Here both projects start from from model levels, though the first step in raider is to grid consistently in the vertical (we had some default values here). Stilt, this is configurable in Raider. So if we used the same setting then that is fine, different settings would lead to slightly different  interpolations/ integrations differences.  

The final step in raider is the 3D cube generation, there there are definitely differences from interpolations
1) The 3D cube in the products: the projection, lateral grid spacing (potentially vertical spacing) are different for opera and nisar.
2) The extraction for 2D grids, for the OPERA is flown through its own intersection code with the DEM. Bilinear lateral and linear vertical. That could be another source for interpolation differences being introduced if that differs from the gunw flow.

Tagging @mgovorcin as he knows the under the hood details on the creation of the opera ztd product.

hfattahi:
> hi @hfattahi, thats great. I had not compared it yet with the tropo within the GUNW, though its on the list. This looks good and indeed this is within the model uncertainties. I suspect this is pretty much related to interpolation differences. A past study from ECMWF for ESA (will see if i can dig up this report) showed that interpolation introduced like to 10% delay variations when they were comparing delays computed starting from pressure and model levels of the same model.
> 
> Here both projects start from from model levels, though the first step in raider is to grid consistently in the vertical (we had some default values here). Stilt, this is configurable in Raider. So if we used the same setting then that is fine, different settings would lead to slightly different interpolations/ integrations differences.
> 
> The final step in raider is the 3D cube generation, there there are definitely differences from interpolations
> 
> 1. The 3D cube in the products: the projection, lateral grid spacing (potentially vertical spacing) are different for opera and nisar.
> 2. The extraction for 2D grids, for the OPERA is flown through its own intersection code with the DEM. Bilinear lateral and linear vertical. That could be another source for interpolation differences being introduced if that differs from the gunw flow.
> 
> Tagging @mgovorcin as he knows the under the hood details on the creation of the opera ztd product.

Thanks for the explanation. It makes sense.

hfattahi (approved):
LGTM. Thanks for the addition.

hfattahi:
@dbekaert I was about to merge this. But I see that @yunjunz just requested an additional review. I will wait for @huchangyang and hope to merge by end of the week if no major comments.

src/mintpy/tropo_opera.py
```diff
@@ -0,0 +1,921 @@
...
+def get_geom_lat_lon_bounds(geom_file):
+    """Get geometry file lat/lon bounds in degrees."""
+    from mintpy.tropo_pyaps3 import get_bounding_box
+
+    meta = readfile.read_attribute(geom_file)
+    lat0, lat1, lon0, lon1 = get_bounding_box(meta, geom_file=geom_file)
+    return min(lat0, lat1), max(lat0, lat1), min(lon0, lon1), max(lon0, lon1)
+
+
+def get_opera_crop_indices(lat, lon, geom_file, pad_cells=3):
```
scottstanie, line 53: `pad_cells` isn't used; it's hardcoded to 0.2 below

src/mintpy/tropo_opera.py
```diff
@@ -0,0 +1,921 @@
...
+        # If no pixels in the entire image live at this altitude, skip the layer entirely!
+        if not np.any(mask_any):
+            continue
+
+        # Interpolate ONLY for the required pixels
+        coords_k = np.vstack([y_idx[mask_any], x_idx[mask_any]])
+        delay_k = map_coordinates(
+            data[k, :, :],
+            coords_k,
+            order=3,
```
scottstanie, line 257: if we are doing order-3, we might wanna bump up the hard coded 0.2 degree buffer. i think that might erode some data at the ends in some cases (i thought i remember changing the crop to more like 2-3 degrees around the request AOI, since the netcdfs are so coarse anyway)

src/mintpy/tropo_opera.py
```diff
@@ -0,0 +1,921 @@
...
+    mobj = re.search(
+        r'OPERA_L4_TROPO-ZENITH_(\d{8}T\d{6}Z)_(\d{8}T\d{6}Z)_HRES_v',
+        text,
+    )
+    if not mobj:
+        return None, None
+    return mobj.group(1), mobj.group(2)
+
+
+def _create_asf_session(asf):
```
scottstanie, line 406: more of a yunjun opinion, but should this be in some other module rather than tucked into one variant of atmospheric correction?

src/mintpy/tropo_opera.py
yunjunz: Thank you Scott for pointing this out. Here is the difference between using a 3-degree buffer and using a 0.2-degree buffer, very small. But as Scott said, the file is small anyway, I would think changing it to 3-degree is a good idea. 

OPERA_DIFF

In the future, we could save all downloaded OPERA TROPO products into the `$WEATHER_DIR`, similar to ERA5, then they can be shared easily among different datasets. For that, we would need a bigger area as well. For PyAPS, I used 10-degree:
https://github.com/insarlab/MintPy/blob/1354ec1aeaf2c1149175d92d2a6da4b83f21ea50/src/mintpy/tropo_pyaps3.py#L291

yunjunz:
Hi @dbekaert and @hfattahi, I have tested the PR, and the code runs fine without error. I have two issues:

1. Comparison between OPERA TROPO and ERA5 from PyAPS based on the [ARIA example dataset over the San Francisco Bay area](https://github.com/insarlab/MintPy/blob/main/docs/demo_dataset.md#sentinel-1-on-san-francisco-bay-with-aria) shows quite different results as below, unlike what you showed above.

+ Double-differenced tropo delay from OPERA TROPO
OPERA_wrap4

+ Double-differenced tropo delay from ERA5
ERA5_wrap4

Here is how I ran:
```bash
tropo_opera.py -f timeseries.h5 -g inputs/geometryGeo.h5
view.py inputs/OPERA.h5 --ref-date 20200110 --ref-yx 200 300 --wrap --wrap-range -2 2

view.py inputs/ERA5.h5 2020 --ref-date 20200110 --ref-yx 200 300 --wrap --wrap-range -2 2 
```

Did I miss anything, or is this centimeter scale difference expected?

2. The speed of data downloading from ASF is fine (for 104 acquisitions, it took less than 20 mins for me), the computation (mainly the 2d/1d interpolation I guess) is very fast (in seconds) for short time series (5 acqusitions), but super slow (estimated to be ~11 hours) for long time series (104 acquisitions), as shown below. Any idea why?

```bash
output file: inputs/OPERA.h5
1) output file either does NOT exist or is NOT newer than all OPERA files.
run or skip: run
--------------------------------------------------
create HDF5 file: inputs/OPERA.h5 with w mode
create dataset  : date       of |S8                       in size of (104,)               with compression = None
create dataset  : timeseries of <class 'numpy.float32'>   in size of (104, 510, 510)      with compression = None
close  HDF5 file: inputs/OPERA.h5
read incidenceAngle from file: inputs/geometryGeo.h5
[==>                      6%                       ] OPERA_L4_TROPO-ZENITH_20160506T120000Z_20251119T065919Z_HRES_v1.0.nc 2572s / 40302s
```

yunjunz:
@dbekaert Another question regarding the OPERA TROPO products, but not related to the code here: I see that this product availability starts from 2016-01-01, but Sentinel-1 products in many areas are earlier than this. Is there a plan to extend the availability of OPERA TROPO products in the future?

yunjunz:
Here is another comparison based on the [Fernandina example dataset](https://github.com/insarlab/MintPy/blob/main/docs/demo_dataset.md#sentinel-1-on-fernandina-with-isce2topsstack).

+ Double-differenced tropo delay from OPERA TROPO
OPERA_wrap4

+ Double-differenced tropo delay from ERA5
ERA5_wrap4

dbekaert:
hi @yunjunz, i missed the replies hence a bit slower to respond here.

Just to note that the figures in this PR from @hfattahi are comparing the same model but implemented as within the NISAR product and those created by OPERA. Here we would expect minimal differences.

In terms of the ERA5 comparison. 
- I did do that in my initial PR https://github.com/insarlab/MintPy/pull/1473. I think we should probably do a comparison more towards GACOS is you want to compare at min the same input model through a different implementation.
- ERA5 has a coarse resolution so should typically be a lot smoother than HRES. Potentially looking at a larger scale would at min demonstrate the models to behave similar. Locally i would expect the higher resolution and dynamics to allow for larger signals that remain to be a model.  The SF Bay area i can imagine that there are much more localized weather effects than the ERA5 model could show. E.g. the incoming fogs that would be predominately on the west side.

dbekaert:
> 3\. The speed of data downloading from ASF is fine (for 104 acquisitions, it took less than 20 mins for me), the computation (mainly the 2d/1d interpolation I guess) is very fast (in seconds) for short time series (5 acqusitions), but super slow (estimated to be ~11 hours) for long time series (104 acquisitions), as shown below. Any idea why?

@hfattahi @scottstanie  any thoughts? I did not bump into this. do you see where it might degrade? I guess one thing is that the intersection or the weight of the nodes does not need to change as long as the grid does not change. so there could be the case to just compute weights ones and then do some sort of element wise operation. That would even be faster than making the interpolation objects over-again.

sssangha:
Hello @yunjunz , circling back on this, is this ready to merge on you end?

Ideally we can integrate it in time for the EarthScope course. Thanks!

yunjunz:
Thank you @dbekaert for the explanation and @sssangha for the reminder. The PR looks good to me to be merged. 

As for the speed, we could improve it later in another PR.

PR #1487, opened 2026-04-06
Add NISAR loading and auxiliary stack support

ehavazli:
## Summary

This PR improves the NISAR data-loading workflow in MintPy by tightening input validation in `load_data.py` and expanding `prep_nisar.py` to generate additional derived products.

## What changed

- require a real DEM path for `processor=nisar`
- fail early with clear errors when the DEM or input GUNW files are missing
- only pass a water mask when it is explicitly set and exists
- add support for auxiliary NISAR outputs:
  - ionosphere stack
  - troposphere stack
  - solid Earth tides (SET) stack
  - standalone water mask
- align DEM and mask warping to the exact NISAR grid
- improve geometry interpolation by evaluating only valid unwrapped pixels
- harden subset, raster-bound, and date-pair handling

## Why

These changes make the NISAR ingestion path more robust and more complete. They reduce configuration errors, improve failure messages, and produce a richer set of MintPy-ready outputs for downstream analysis.

## Testing

- validated input checks for missing/invalid DEM and GUNW paths
- verified `prep_nisar.py` writes geometry, interferogram, ionosphere, troposphere, SET, and water mask outputs with the updated workflow

## Summary by Sourcery

Tighten NISAR ingestion by validating DEM/GUNW inputs, aligning DEM/mask warping to the NISAR grid, and generating additional NISAR-derived MintPy stacks and water masks.

New Features:
- Generate separate NISAR ionosphere, troposphere, and solid Earth tides stacks alongside the main interferogram stack.
- Create a standalone MintPy-formatted water mask aligned to the NISAR grid and constrained to valid pixels.

Bug Fixes:
- Ensure NISAR processing requires a real, existing DEM file and fails early if DEM or GUNW inputs are missing.
- Compute raster subset bounds and metadata more robustly to avoid empty or invalid subsetting regions.
- Use finite unwrapped phase and fill values to define valid pixels consistently across geometry and masking.

Enhancements:
- Refine DEM and mask warping to match the NISAR grid exactly using EPSG-aware bounds and resolutions.
- Improve radar-grid interpolation of geometry, troposphere, and solid Earth tides by evaluating only at valid unwrapped pixels and handling coordinate ordering issues.
- Extend metadata extraction and subset handling to support polarization-specific datasets and stricter numeric typing.


ehavazli:
@sourcery-ai review

src/mintpy/prep_nisar.py
```diff
@@ -420,29 +1060,25 @@ def prepare_stack(
         "connectComponent": [np.float32, (num_pair, rows, cols), None],
     }
 
-    # initiate HDF5 file
     meta["FILE_TYPE"] = "ifgramStack"
```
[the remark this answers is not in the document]
ehavazli: @yunjunz
"ifgramStack" file type is chosen because of the predefined data types in MintPy. I don't think it will be helpful downstream if I change it over here.

src/mintpy/load_data.py
```diff
@@ -637,10 +637,27 @@ def prepare_metadata(iDict):
...
+            raise ValueError(
+                'mintpy.load.demFile is required for processor=nisar. '
+                'Please set it to a real DEM path in the template.'
+            )
+        dem_files = glob.glob(str(dem_file))
+        if len(dem_files) == 0:
+            raise FileNotFoundError(
+                f'No DEM file found for mintpy.load.demFile: {dem_file}'
+            )
+        dem_file = dem_files[0]
```
hfattahi: If the user has provided a dem_file, why would you need to still look for the dem_files, and then choose the first in the list. What is the reason the first one is the correct one? This logic does not make sense to me. 
I suggest to trust the user input and avoid looking further

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1044 @@
...
 from scipy.interpolate import RegularGridInterpolator
 
 from mintpy.constants import EARTH_RADIUS, SPEED_OF_LIGHT
-from mintpy.utils import ptime, writefile
+from mintpy.utils import attribute as attr, ptime, writefile
+
+# ---------------------------------------------------------------------
+# Constants / HDF5 paths (GUNW frequencyA, unwrappedInterferogram)
+# ---------------------------------------------------------------------
+DATASET_ROOT_UNW = "/science/LSAR/GUNW/grids/frequencyA/unwrappedInterferogram"
```
hfattahi: I understand that we usually have frequencyA. But it would have been nice to not limit the user to frequesnyA. In theory one can run GUNW for frequencyB only and have a stack of frequency B interferograms and they should be still able to run mintpy.

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1044 @@
...
-    'cor'              : f"{DATASET_ROOT_UNW}/POL/coherenceMagnitude",
-    'connComp'         : f"{DATASET_ROOT_UNW}/POL/connectedComponents",
-    #'mask'             : f"{DATASET_ROOT_UNW}/mask",
-    'epsg'             : f"{DATASET_ROOT_UNW}/projection",
-    'xSpacing'         : f"{DATASET_ROOT_UNW}/xCoordinateSpacing",
-    'ySpacing'         : f"{DATASET_ROOT_UNW}/yCoordinateSpacing",
-    'polarization'     : "/science/LSAR/GUNW/grids/frequencyA/listOfPolarizations",
-    'range_look'       : f"{PARAMETERS}/numberOfRangeLooks",
-    'azimuth_look'     : f"{PARAMETERS}/numberOfAzimuthLooks",
+    "xcoord": f"{DATASET_ROOT_UNW}/POL/xCoordinates",
```
hfattahi: Not related to this PR but I'm not sure how POL is handled.

Alex-Lewandowski:
I ran this on a non-redundant stack of 4 nearest-neighbor GUNWs and confirmed that the dataset was loaded. It produced the expected ifgramStack.h5 and geometryGeo.h5, as well as auxiliary outputs including ionStack.h5, setStack.h5, and tropoStack.h5.

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1044 @@
...
         demFile=inps.dem_file,
-        maskFile=inps.mask_file
+        maskFile=inps.mask_file,
+        polarization=pol,
+    )
+
+    # standalone water mask (MintPy format)
+    if getattr(inps, "mask_file", None) not in [None, "None", "auto"]:
+        water_mask_file = os.path.join(inps.out_dir, "waterMask.h5")
+        prepare_water_mask(
```
hfattahi: I don't see the mask dataset in NISAR GUNW been used. I was wondering if there is a reason to not use this dataset which exists in this path:
```
"science/LSAR/GUNW/grids/frequencyA/unwrappedInterferogram/mask"
```
The description of this mask based on GUNW spec is the following:

```
Combination of water mask and a mask of subswaths of valid samples in the reference RSLC and geometrically-coregistered secondary RSLC. Each pixel value is a three-digit number: the most significant digit represents the water flag of that pixel the reference RSLC, where 1 is water and 0 is non-water; the second digit represents the subswath number of that pixel in the reference RSLC; the least significant digit represents the subswath number of that pixel in the secondary RSLC. A value of 0 in either subswath digit indicates an invalid sample in the corresponding RSLC
Can you write a python script to interpret the mask
```
As you see the mask is not only water mask but a combination of water mask and subswath mask for reference and secondary RSLCs. The subswath mask is needed to mask out edges of the swath. while for dithered PRF we have one subswath, the mask nicely identifies the valid region of multiple subswaths in constant PRF. 

assuming the mask dataset in GUNW, you can extract different binary masks as 

```
    water_mask = (mask // 100) == 1

    ref_subswath = (mask // 10) % 10
    
    sec_subswath = mask % 10
    
    is_valid = (ref_subswath > 0) & (sec_subswath > 0)

is_land_and_valid = (is_valid) & ~water_mask

```

Here is a plot of the mask dataset for a frame at coastal region (LA):

Image

And here is a plot of valid region over land (i.e., is_land_and_valid in the code snippet above)

Image

src/mintpy/prep_nisar.py
hfattahi: Found my answer in _datasets_for_pol function. It is fine.

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1044 @@
...
+        src_path=dem_file,
+        src_epsg=dem_src_epsg,
+        dst_epsg=dst_epsg,
+        xcoord=xcoord,
+        ycoord=ycoord,
+        resample_alg="bilinear",
+    )
+
+    y_2d, x_2d = np.meshgrid(ycoord, xcoord, indexing="ij")
+    valid_mask = _read_valid_unw_mask(gunw_file, xybbox, polarization)
```
hfattahi: See my comment below about mask dataset in GUNW

hfattahi (commented):
Thanks @ehavazli for this PR. Exciting to see these additions to make minpy consistent with NISAR GUNW. I have some minor comments. Perhaps my main comment is about the mask dataset in GUNW.

ehavazli:
@hfattahi 
Thanks for the comments. I pushed updates addressing the requested changes.

Summary of changes:

- Added NISAR frequency selection via `-freq/--frequency` with choices `auto`, `A`, and `B`.
- Added `mintpy.load.frequency = auto` to the template and threaded it through `load_data.py`.
- `auto` keeps the current behavior and defaults to `frequencyA`; explicit `--frequency B` is honored and fails clearly if frequency B is missing.
- Refactored the GUNW dataset paths so `prep_nisar.py` no longer hard-codes `frequencyA`.
- Updated `load_data.py` to trust the user-provided DEM path instead of globbing and selecting the first match.
- Added use of the native GUNW mask at `/science/LSAR/GUNW/grids/<frequency>/unwrappedInterferogram/mask`.
- The native mask is decoded into `is_land_and_valid` using water flag plus reference/secondary subswath validity.
- The decoded mask is now written into `geometryGeo.h5` as `waterMask` and also into standalone `waterMask.h5`; if an external mask is supplied, it is intersected with the native mask.
- Fixed DEM warping to preserve the exact target NISAR grid size, avoiding off-by-one shape mismatches.
- Fixed date-pair extraction for generic filenames like `GUNW_product.h5` by reading NISAR identification metadata first.
- Added skip behavior for optional ionosphere, troposphere, and SET stacks when their source layers are missing, with clear warnings for each missing layer.
- Required interferogram layers still fail fast if missing.

Responses to specific review items:

- DEM handling: We now trust the explicit `demFile` path and avoid globbing/selecting the first matching DEM.
- Frequency A/B support: Users can now select frequency A or B; default `auto` preserves frequency A behavior.
- Native GUNW mask: We now use the GUNW mask dataset without a polarization group and decode it into `is_land_and_valid`.

- POL handling: this continues to be handled through `_datasets_for_pol`, which builds polarization-specific paths where the GUNW layout requires POL.
- `osr` import/runtime risk: addressed by importing `osr` explicitly from `osgeo`.
- `FILE_TYPE`: intentionally left as `ifgramStack` because MintPy has predefined stack data types and downstream code expects that convention for these stack-like products.

Testing performed:

- Tested `prep_nisar.py` on the sample GUNW product with DEM.
- Confirmed the workflow writes geometry, water mask, ifgram stack, ion stack, and SET stack.
- Confirmed missing correction layers are skipped with explicit warnings instead of raising a `KeyError`.

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1221 @@
...
+    """Decode the native GUNW mask into MintPy's keep-mask convention."""
+    datasets = _datasets_for_pol(pol, frequency)
+    path = datasets["mask"]
+
+    with h5py.File(gunw_file, "r") as ds:
+        if path not in ds:
+            return _read_unwrapped_phase_valid_mask(gunw_file, xybbox, pol, frequency)
+
+        dset = ds[path]
+        mask = dset[xybbox[1] : xybbox[3], xybbox[0] : xybbox[2]]
```
hfattahi, line 282: Thanks for adding the mask reader. I think this is fine for the current version of GUNW products. However in a recent ISCE3 PR [here](https://github.com/isce-framework/isce3/pull/247) we are modifying the mask. With the upcoming release (X05014) the mask will chage the data type from 8 bit Unsigned to 32 bits Unsigned Int and have the following description:

```
Combination of a water mask, a mask of subswaths of valid samples, and data anomalies"
                 " in the reference RSLC and the geometrically coregistered secondary RSLC."
                 " Each pixel value is encoded as a 32-bit unsigned integer."
                 " Bits 0–7 represent subswath encoding, where the most significant digit represents"
                 " the water flag of that pixel in the reference RSLC, where 1 is water"
                 " and 0 is non-water; the second most significant digit corresponds to"
                 " the subswath number of the reference RSLC, and the least significant digit"
                 " corresponds to the subswath number of the secondary RSLC;"
                 " a value of 0 in either digit indicates an invalid sample in the corresponding RSLC."
                 " Bits 8–15 represent bitwise anomaly flags for the secondary RSLC, and"
                 " bits 16–23 represent bitwise anomaly flags for the reference RSLC,"
                 " with each bit corresponding to a specific anomaly condition."
                 " A value of 0 in the anomaly bits indicates that no anomaly is detected in the corresponding RSLC."
                 " Bits 24–31 are reserved for future use"
```
As you can see the first 8 bits of the new 32 bits mask is identical to the current 8 bits mask. So with a very minor change you can keep your reader as is for now. In particular you can simply modify the following to extract the first 8 bits which contains water and subswath masks:

```suggestion
        mask_bits = dset[xybbox[1] : xybbox[3], xybbox[0] : xybbox[2]]
        # Bits 0–7: Subswath and Water encoding
        mask = mask_bits & 0xFF
```
With this change your reader should work for current and future GUNW.

hfattahi (commented):
Thanks @ehavazli . The changes look good to me. I have one more comment on the mask. Please see details below.
Also one question for you. I obviously have forgotten when in the pipeline an "average spatial coherence" is written to disk. I managed to run your PR but I did not see any avgSpatialCoherence.h5 being generated.

src/mintpy/prep_nisar.py
```diff
@@ -11,406 +13,1221 @@
...
+        externalMaskFile=inps.mask_file,
+        polarization=pol,
+        frequency=frequency,
+    )
 
+    # standalone water mask (MintPy format)
+    water_mask_file = os.path.join(inps.out_dir, "waterMask.h5")
+    prepare_water_mask(
+        outfile=water_mask_file,
+        metaFile=input_files[0],
```
hfattahi, line 566: @ehavazli it looks to me that the valid data mask is read only for the first interferogram. This can be problematic. The valid data mask may be slightly different between different acquisitions and therefore may be different between different GUNWs.

ehavazli:
@hfattahi
The `avgSpatialCoh.h5` is produced during the reference point selection with `temporal_average.py`. I made the changes you requested, please see a summary of changes below.

Summary of changes:

- The native GUNW mask reader now extracts only bits 0-7 using mask_bits & 0xFF, so it remains compatible with both current 8-bit masks and upcoming 32-bit masks.
- _FillValue is checked on the raw mask values before bit extraction, preventing fill/no-data pixels from being interpreted as valid after bit masking.
- The valid-land mask is now computed across all input GUNW products, not just the first one.
- Per-product masks are combined with a logical AND, so any pixel masked in one GUNW is masked consistently across the full stack.
- The resulting common mask is used for geometryGeo.h5, standalone waterMask.h5, and all generated stack products.
- Optional external masks are still supported and are applied as an additional refinement to the common GUNW-derived mask.

hfattahi (approved):
Thank you @ehavazli for addressing my comments. Highly appreciate it. LGTM!

PR #1490, opened 2026-05-06
Add opt-in torch GPU solver for invert_network

### LINKED ISSUE #1489: RFC: opt-in GPU backend for invert_network (torch.linalg.lstsq, CUDA)
# RFC: opt-in GPU backend for invert_network (torch.linalg.lstsq, CUDA)

## Motivation

`invert_network` is consistently the longest single step in time-series InSAR processing. Within `run_ifgram_inversion_patch`, the dominant cost on real data is **per-pixel Python looping**.

The current CPU path splits pixels into three cases:

1. OLS pixels with all ifgrams valid → solved in a single `scipy.linalg.lstsq` call (good).
2. OLS pixels with per-pixel NaN masks → **Python loop, one `scipy.linalg.lstsq` call per pixel**.
3. WLS pixels (per-pixel weights) → **Python loop, one `scipy.linalg.lstsq` call per pixel**, regardless of NaN pattern.

On real data, NaN observations (atmosphere, decorrelation) are common and WLS is preferred for accuracy, so cases 2 and 3 dominate. `scipy.linalg.lstsq` cannot vectorize these because pixels effectively have different design matrices (different row masks, different weights).

`torch.linalg.lstsq` accepts a **batched** stack of independent (A_k, y_k) systems and dispatches to CUDA. This eliminates the Python loop in cases 2 and 3 by solving them in a single GPU call.

## Proposal

Add an opt-in **CUDA-only** GPU backend for `invert_network`, selectable via template (or `--backend` CLI flag):

    mintpy.networkInversion.backend = auto | cpu | torch

- `auto` (default): resolves to `cpu` via the existing `check_template_auto_value` static lookup. Behavior **unchanged**
  for any user who does not modify their template.
- `cpu` (explicit): existing scipy path, byte-for-byte unchanged.
- `torch` (explicit opt-in): batched `torch.linalg.lstsq` on CUDA.

PyTorch is gated behind a new `[gpu]` extras group in `pyproject.toml`. Users without the extras pay zero — neither dependency nor runtime cost.

### Scope (intentionally narrow)

- **CUDA only.** When the user requests `backend='torch'` and CUDA is unavailable, the run **fails fast with an explicit error** rather than silently falling back. Rationale: the user explicitly opted in; silent fallback would mask configuration / driver issues. Users without CUDA simply leave `backend = auto` (default cpu).
- **No CPU torch backend.** Out of scope. Could be a separate proposal later if there is demand; would expand test surface and maintenance.
- **Full-rank pixels only on GPU.** CUDA's `gels` driver does not handle rank-deficient systems. Rare in real SBAS networks;
  encountered cases produce NaN in the output. CPU path retains its existing rank handling.

## Evidence (reference run, single machine)

On FernandinaSenDT128 with an RTX 5080 (16 GiB VRAM, warm SSD) I observed **1.43× wall-time speedup** for the `invert_network` step vs the CPU path. Numerical equivalence verified at float32 round-off. A modest figure on a tutorial dataset, but applied to production-scale runs where `invert_network` dominates wall, the absolute time saved is practical. A larger-scene benchmark is planned to confirm the scaling story before PR.

- Working prototype:
  https://github.com/s-sasaki-earthsea-wizard/MintPy

## Open questions for maintainers

1. **API surface**: template flag (current) vs CLI flag vs env var — preference?
2. **Extras layout**: `[gpu]` (current) vs more specific name (`[cuda]` / `[torch-cuda]`)?
3. **CI**: comfortable adding a GPU-tagged CI job (or keeping it manual)? A no-op smoke import test is cheap and would catch packaging regressions.
4. **Docs scope**: install steps in existing `docs/installation.md`, or a separate `docs/gpu.md` modelled on `docs/dask.md`?

Following the Dask integration playbook (#349 → #351 → #357) as a reference for staged contribution. Happy to split into smaller issues or PRs as preferred.

s-sasaki-earthsea-wizard:
## Description of proposed changes

This PR adds an opt-in CUDA-accelerated path for the per-pixel weighted
least-squares inversion in the `invert_network` step (`ifgram_inversion.py`).
The fork has been running this code on tutorial-scale and large-scale scenes
for several weeks; this submission consolidates the implementation as it
currently stands.

The default path is unchanged. `mintpy.networkInversion.solver = auto`
resolves to `cpu`, and the existing CPU code path is byte-for-byte identical
to upstream — every other step in `smallbaselineApp.py` continues to run on
the CPU regardless of this setting.

The aim is to contribute faster InSAR time-series processing for NVIDIA GPU
users, since `invert_network` is the dominant CPU bottleneck on typical
workflows and the gap widens with scene size.

Closes #1489 (RFC).

### Implementation summary

- New module `src/mintpy/ifgram_inversion_gpu.py` batches the per-pixel WLS
  systems on a single CUDA device. The solver is **normal-equations +
  Cholesky** via `torch.linalg.cholesky_ex`, which (a) is significantly
  faster than `torch.linalg.lstsq` on the matrix shapes encountered here and
  (b) lets us detect rank-deficient pixels through the returned `info` codes
  rather than via post-hoc residual checks.
- `ifgram_inversion.py` dispatches to the GPU module only when
  `solver = torch` is explicitly requested; the CPU loop is untouched.
- New `[gpu]` extras in `pyproject.toml`, sourced from `requirements-gpu.txt`
  (just `torch>=2.11`). Install requires the PyTorch CUDA wheel index:
  `pip install -e ".[gpu]" --extra-index-url https://download.pytorch.org/whl/cu128`
  (documented in `docs/installation.md` §2.4).
- Tests in `tests/test_ifgram_inversion_gpu.py` cover the dispatch logic and
  the GPU fast paths with synthetic NaN / rank-deficient fixtures.

### Behavior notes

- **VRAM auto-sizing** — `gpuChunkSize = 0` (the default) probes free GPU
  memory at runtime and chooses a per-chunk pixel count with a fixed
  headroom factor; passing a positive integer overrides this for
  reproducible chunking across hosts with different VRAM.
- **Rank-deficient pixels** are detected via `cholesky_ex` info codes and
  zeroed so NaN/Inf cannot propagate downstream; a warning line reports
  the count per chunk.
- **Per-pixel NaN observations** are handled by zeroing the corresponding
  row weight, which is mathematically equivalent to dropping that row from
  the WLS system.
- **No silent CPU fallback** — selecting `solver = torch` on a host without
  a visible CUDA device raises immediately rather than silently falling
  back to CPU; this keeps any performance regression visible.

### Design pivot vs the original RFC

The RFC (#1489) originally described `torch.linalg.lstsq` as the GPU solver.
During development the path was switched to normal-equations + Cholesky
after a side-by-side benchmark showed it preserves output equivalence to
float32 round-off (RMS ~1e-5) while running ~16× faster than `lstsq` on the
same matrix shapes (tutorial dataset: FernandinaSenDT128). An RMS difference
on the order of `1e-5` in the displacement field is well below the typical
InSAR noise floor — sub-millimeter on a per-pixel basis — so the two solvers
are operationally equivalent for the geophysical use case. The `lstsq` path
was removed before this submission so there is only one supported GPU code
path to reason about.

### Performance

Indicative numbers measured on an NVIDIA RTX 5080 (Blackwell sm_120, CUDA
12.8, PyTorch 2.11). Speedup will vary with scene size, GPU class, and
chunk-size tuning.

| Scene | Pixels | ifgs | `invert_network` internal | step wall |
|---|---:|---:|---:|---:|
| FernandinaSenDT128 (tutorial) | 270k | 288 | ~16× faster | ~4.5× faster |
| GalapagosSenDT128 (large) | 3.4M | 475 | ~44× faster | ~36× faster |

Large-scene absolute timings: CPU 6189 s → torch 170 s on the same machine.
Numerical equivalence between the `cpu` and `torch` solvers holds to
float32 round-off in both cases (RMS on the order of `1e-5`; absolute RMS
max ~16 µm on the large-scene case).

Reproduction artifacts (harness scripts, raw logs, full reports) live in a
separate repository. Links below are pinned to a single sibling commit so
the data does not move during review:

- `cpu` vs `torch` end-to-end on Fernandina:
  <https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/c20ca8bb/reports/report_torch.md>
- `lstsq` vs Cholesky equivalence + per-step speedup:
  <https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/c20ca8bb/reports/report_solver_comparison.md>
- Chunk-size sensitivity sweep:
  <https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/c20ca8bb/reports/report_chunk_sweep.md>
- `torch.profiler` GPU kernel breakdown:
  <https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/c20ca8bb/reports/report_profile.md>
- Large-scene Galapagos run:
  <https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/c20ca8bb/reports/report_large_scene.md>

Numbers are from a single development machine; absolute timings will vary
across hardware, but the qualitative findings (Cholesky > lstsq; GPU > CPU
at this matrix scale; speedup grows with scene size) should hold for any
recent NVIDIA CUDA-class device. Harness scripts and raw logs in the
mintpy-benchmark repository above let other GPU users reproduce on their
own data.

### Local validation

Run on the PR branch (`upstream/main` + the three commits in this PR), against
FernandinaSenDT128:

- **`pre-commit run --all-files`** exits clean (13 hooks pass + 1 skip on json,
  per the upstream `.pre-commit-config.yaml`).
- **`smallbaselineApp.py` end-to-end with default settings** (`solver = auto`
  resolves to `cpu`): all 18 steps, `Normal end of smallbaselineApp processing!`,
  total wall time 2 h 2 m (the `correct_troposphere` step's CDS download
  dominated; the actual computation portion is small). All standard output
  products generated (`timeseries.h5`, `timeseries_ERA5.h5`,
  `timeseries_ERA5_ramp.h5`, `timeseries_ERA5_ramp_demErr.h5`, `velocity.h5`,
  `velocityERA5.h5`, `geo/`, etc.).
- **`smallbaselineApp.py` end-to-end with `solver = torch`** (same dataset,
  ERA5 grib + tropo product reused via symlink so this run only re-exercises
  `invert_network` and the post-tropo steps): all 18 steps, `Normal end of
  smallbaselineApp processing!`, total wall time 6 m 57 s. Log confirms the
  GPU path was actually entered:

  ```
  mintpy.networkInversion.solver: auto --> torch
  estimating time-series via torch solver (batched, GPU)
  GPU auto chunk_size = 19403 pixels (free VRAM 15.1 GiB)
  estimating time-series via torch batched WLS in 14 chunk(s) of up to 19403 pixels ...
  ```

  Same set of standard output products as the CPU run.

### Disclosure

This work was developed with the assistance of Claude Opus 4.7 (Anthropic's coding
assistant). All design decisions, benchmark execution, and review of the
generated code were performed by me. Per project convention the
`Assisted-by: Claude Opus 4.7` trailers used during fork development have been stripped
from this branch's commit history; this paragraph is the canonical
disclosure.

If the AI-assisted aspect raises review or maintenance concerns for the
project, I'm happy to discuss — including whether to keep the GPU module
opt-in / under a feature flag.

## Reminders

- [x] Fix #1489
- [x] Pass Pre-commit check (green) — verified locally with the upstream `.pre-commit-config.yaml`; CI to confirm.

## Summary by Sourcery

Add an opt-in PyTorch CUDA path to accelerate the invert_network least-squares inversion while leaving default CPU processing unchanged.

New Features:
- Add an opt-in PyTorch CUDA solver for batched weighted and unweighted network inversion with configurable GPU chunk sizing.
- Expose solver and GPU chunk-size controls through the CLI, templates, and network inversion configuration.

Enhancements:
- Preserve the existing CPU solver as the default while providing CUDA availability checks, NaN handling, rank-deficiency protection, and GPU memory-based chunk sizing.

Build:
- Add an optional [gpu] dependency extra for CUDA-enabled PyTorch installations.

Documentation:
- Document GPU installation, configuration, usage, behavior, and performance alongside the existing Dask documentation.

Tests:
- Add CUDA-gated tests covering numerical equivalence, weighted and unweighted inversion, NaN observations, phase inversion, chunk-size consistency, and unsupported solver validation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -88,6 +88,8 @@ Algorithms implemented in the software are described in details at [Yunjun et al
 + [Example data directory](./dir_structure.md)
 + [Example template files](./templates/README.md)
 + [Tutorials in Jupyter Notebook](https://github.com/insarlab/MintPy-tutorial)
++ [Parallel processing with Dask](./dask.md)
++ [GPU acceleration for the `invert_network` step (opt-in PyTorch CUDA solver, partial)](./gpu.md)
 
 ### 4. Contact us
 
--- a/docs/dask.md
+++ b/docs/dask.md
@@ -7,6 +7,8 @@ Most computations in MintPy are operated in either a pixel-by-pixel or a epoch-b
 
 [Here](https://github.com/2gotgrossman/dask-rsmas-presentation) is an entry-level presentation on parallel computing using Dask by David Grossman. Below we brief describe for each cluster/scheduler the required options and recommended best practices.
 
+For GPU acceleration of the `invert_network` step on a single CUDA device — orthogonal to the Dask paths described here — see [gpu.md](./gpu.md).
+
 ## 1. local cluster ##
 
 The parallel processing on a single machine is supported via [`Dask.distributed.LocalCluster`](https://docs.dask.org/en/latest/setup/single-distributed.html#localcluster). This is recommended if you are running MintPy on a local machine with multiple available cores, or on an HPC but wish to allocate only a single node's worth of resources.
--- a/docs/gpu.md
+++ b/docs/gpu.md
@@ -1,0 +1,68 @@
+# Configure GPU acceleration for the network inversion #
+
+The `invert_network` step (in `ifgram_inversion.py`) ships an opt-in GPU solver that batches the per-pixel weighted least-squares inversion as normal-equations + Cholesky on a CUDA device via PyTorch. This is a partial GPU implementation: only `invert_network` is offloaded to the GPU; every other step in `smallbaselineApp.py` continues to run on the CPU. The solver is opt-in — the default `mintpy.networkInversion.solver = auto` resolves to `cpu`, so existing setups are unaffected.
+
+The `torch` solver is orthogonal to Dask parallel processing (see [dask.md](./dask.md)): the former replaces the per-pixel CPU loop with a single batched Cholesky on one CUDA device, the latter distributes that same per-pixel loop across multiple worker processes. The two paths are not currently combined; pick one.
+
+## 1. Setup ##
+
+See [installation.md](./installation.md) section 2.4 for installing the `[gpu]` extras with the matching CUDA wheel index. Selecting `solver = torch` on a host without a visible CUDA device is a hard error (no silent CPU fallback).
+
+## 2. Enable ##
+
+#### 2.1 via command line ####
+
+Run the following in the terminal:
+
+```bash
+ifgram_inversion.py inputs/ifgramStack.h5 --solver torch
+ifgram_inversion.py inputs/ifgramStack.h5 --solver torch --gpu-chunk-size 20000
+```
+
+`--gpu-chunk-size 0` (the default) auto-sizes the per-chunk pixel count from free VRAM; pass a positive integer to override.
+
+#### 2.2 via template file ####
+
+Adjust options in the template file:
+
+```cfg
+mintpy.networkInversion.solver       = torch  #[cpu / torch], auto for cpu
+mintpy.networkInversion.gpuChunkSize = auto   #[int >= 0], auto for 0 (auto-size from free VRAM)
+```
+
+and feed the template file to the script:
+
+```bash
+ifgram_inversion.py inputs/ifgramStack.h5 -t smallbaselineApp.cfg
+smallbaselineApp.py smallbaselineApp.cfg
+```
+
+#### 2.3 Testing using example data ####
+
+Download and run the FernandinaSenDT128 example data; then run with and without the GPU solver:
+
+```bash
+cd FernandinaSenDT128/mintpy
+ifgram_inversion.py inputs/ifgramStack.h5 -w no --solver cpu
+ifgram_inversion.py inputs/ifgramStack.h5 -w no --solver torch
+```
+
+The two outputs should agree to float32 round-off (RMS on the order of 1e-5).
+
+## 3. Behavior notes ##
+
++ **VRAM auto-sizing.** `gpuChunkSize = 0` (auto) probes free VRAM at runtime and chooses a per-chunk pixel count with a fixed headroom factor. Set an explicit integer to override (e.g. for reproducible chunking across hosts with different VRAM).
+
++ **Rank-deficient pixels.** Detected via `torch.linalg.cholesky_ex` info codes; their solution is set to zero so NaN/Inf never propagate downstream. A warning line reports the count per chunk.
+
++ **Per-pixel NaN observations.** Handled by zeroing the corresponding row weight, which is mathematically equivalent to dropping that row from the WLS system.
+
++ **No silent CPU fallback.** Selecting `solver = torch` on a host without a visible CUDA device raises immediately rather than silently falling back to CPU; this keeps performance regressions visible.
+
+## 4. Performance ##
+
+Indicative numbers below were measured on an NVIDIA RTX 5080 (Blackwell sm_120, CUDA 12.8, PyTorch 2.11) at the time this feature was submitted. Speedup depends on scene size, GPU class, and chunk-size tuning, so reproduce on your own data and hardware before drawing conclusions.
+
++ **Tutorial-scale** (FernandinaSenDT128: 270k pixels, 288 ifgs) — `invert_network` runs roughly **16×** faster internally and **4.5×** faster end-to-end versus the CPU path.
++ **Large-scene** (GalapagosSenDT128: 3.4M pixels, 475 ifgs; ~12.6× pixels and 1.65× ifgs over Fernandina) — roughly **44×** internal and **36×** step-wall speedup on `invert_network` (CPU 6189 s → torch 170 s on the same machine), confirming the speedup grows at scale.
++ **Numerical equivalence** between the `cpu` and `torch` solvers holds to float32 round-off: RMS on the order of `1e-5` on the tutorial case, with absolute RMS at most ~16 µm on the large-scene case.
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -194,6 +194,64 @@ Same as the <a href="#21-install-on-linux">instruction for Linux</a>, except for
 </details>
 </p>
 
+### 2.4 Optional: GPU acceleration via PyTorch CUDA ###
+
+<p>
+<details>
+<p><summary>Click to expand for more details</summary></p>
+
+<p>The <code>invert_network</code> step ships an opt-in GPU solver that solves the per-pixel WLS inversion as a batched normal-equation + Cholesky on a CUDA device via PyTorch. It is opt-in: the default <code>mintpy.networkInversion.solver = auto</code> resolves to <code>cpu</code>, so existing setups are unaffected. There is no silent CPU fallback — selecting <code>torch</code> without a visible CUDA device is a hard error.</p>
+
+<h4>a. Prerequisites</h4>
+
+<ul>
+<li>An NVIDIA GPU with a working CUDA driver</li>
+<li>MintPy installed from source in editable mode (Section 2 above)</li>
+</ul>
+
+<h4>b. Install the [gpu] extras</h4>
+
+<p>The <code>[gpu]</code> extras pull a CUDA-enabled PyTorch build. Pick the <code>cuXXX</code> wheel index that matches your CUDA toolkit version (see <a href="https://pytorch.org/get-started/locally/">pytorch.org</a> for the current list); for example, <code>cu121</code>, <code>cu124</code>, or <code>cu128</code>:</p>
+
+```bash
+python -m pip install -e ".[gpu]" \
+    --extra-index-url https://download.pytorch.org/whl/cu128
+```
+
+<p>If you use <a href="https://docs.astral.sh/uv/">uv</a> instead of <code>pip</code>, add <code>--index-strategy unsafe-best-match</code> to work around a stale <code>setuptools</code> pin in the PyTorch wheel index:</p>
+
+```bash
+uv pip install -e ".[gpu]" \
+    --extra-index-url https://download.pytorch.org/whl/cu128 \
+    --index-strategy unsafe-best-match
+```
+
+<h4>c. Verify</h4>
+
+```bash
+python -c "import torch; print(torch.cuda.is_available())"
+# expected: True
+```
+
+<h4>d. Enable</h4>
+
+<p>Set the template flag:</p>
+
+```cfg
+mintpy.networkInversion.solver = torch
+```
+
+<p>or pass it on the command line:</p>
+
+```bash
+ifgram_inversion.py inputs/ifgramStack.h5 --solver torch
+```
+
+<p>See <a href="./gpu.md">gpu.md</a> for tuning, behavior notes, and benchmarks.</p>
+
+</details>
+</p>
+
 ## 3. Post-Installation Setup ##
 
 #### a. ERA5 for tropospheric correction ####
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,6 +114,9 @@ readme = { file = ["docs/README.md"], content-type = "text/markdown" }
 [tool.setuptools.dynamic.optional-dependencies.test]
 file = ["tests/requirements.txt"]
 
+[tool.setuptools.dynamic.optional-dependencies.gpu]
+file = ["requirements-gpu.txt"]
+
 [tool.setuptools.packages.find]
 where = ["src"]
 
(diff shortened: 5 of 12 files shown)


s-sasaki-earthsea-wizard:
Sharing additional benchmark data for the GPU torch solver. The PR description's Performance section reports per-step `invert_network` numbers on FernandinaSenDT128 and GalapagosSenDT128 (both ISCE2 / Sentinel-1 C-band). The bench has since been extended to 5 scenes spanning 4 InSAR processors (ISCE2, GMTSAR, ARIA, ROI_PAC) and 2 sensors / wavelengths (Sentinel-1 C-band, ALOS-1 L-band):

| Scene | Processor / Sensor | Pixels | Ifgs (K) | Dates (D) | `invert_network` cpu wall | torch wall | speedup |
|---|---|---:|---:|---:|---:|---:|---:|
| FernandinaSenDT128 | ISCE2 / S1 (C) | 270 k | 288 | 98 | 645.12 s | 6.88 s | **93.77×** |
| GalapagosSenDT128 | ISCE2 / S1 (C) | 3.40 M | 490 | 98 | 2976.72 s | 79.40 s | **37.49×** |
| SanFranBaySenD42 | GMTSAR / S1 (C) | 326 k | 1297 | 333 | 1080.38 s | 17.42 s | **62.02×** |
| KujuAlosAT422F650 | ROI_PAC / ALOS-1 (L) | 226 k | 167 | 24 | 31.01 s | 4.53 s | **6.85×** |
| SanFranSenDT42 | ARIA / S1 (C) | 1.04 M | 505 | 114 | 58.85 s | 11.07 s | **5.32×** |

Configuration: warm SSD, NVIDIA RTX 5080 (16 GiB), float32, `mintpy.networkInversion.solver = torch`. Each run was end-to-end `smallbaselineApp.py` (full 18-step pipeline), not a direct call to the solver — wall numbers above are extracted from `/usr/bin/time -v` capture of the `invert_network` segment. The cpu-only steps in the same pipeline (`load_data`, `modify_network`, `correct_SET`, `correct_troposphere`, `deramp`, `save_hdfeos5`) stay within ±5 % between the cpu and torch runs of each scene, serving as an I/O / cache control.

The 5 — 94× range is structurally driven by per-pixel solve cost (∝ K · D²): Kuju (K=167, D=24) sits at the floor, SanFranBay (K=1297, D=333) at the ceiling. The Fernandina and Galapagos figures above are consistent with the `~16×` internal / `~4.5×` step-wall (Fernandina) and `~44×` internal / `~36×` step-wall (Galapagos) numbers in the original PR description; the larger headline here reflects the warm-SSD scene root and the per-step wall extracted from a fresh end-to-end run with both cpu and torch using identical fixtures.

Numerical agreement: the float32 round-off gate (rms / |cpu|.max < 1e-5) is met for the user-visible final products (`velocity.h5`, geocoded outputs) in all 5 scenes. Two scenes (Kuju, SanFranSF) show divergence on radar-coordinate intermediate products at rms/scale 1 — 7 %, but Kuju's geocoded velocity (filtered through `maskTempCoh.h5`) passes at 1.38e-7 — consistent with the divergence being confined to pixels that the downstream `maskTempCoh.h5` mask drops anyway. Diagnosed in the report as cpu `scipy.linalg.lstsq` min-norm fill vs torch `cholesky_ex` fill for near-rank-deficient masked-out pixels; making the radar-coord diff tool mask-aware is queued as a sibling-repo follow-up.

Full per-step wall breakdown, fixture parity verification (cpu and torch fixtures verified byte-identical except for the 2 `mintpy.*.solver = torch` lines), and the numerical comparison methodology are in the report:

→ [`reports/report_end_to_end_bench.md` @ `0fbf71b`](https://github.com/s-sasaki-earthsea-wizard/mintpy-benchmark/blob/0fbf71b/reports/report_end_to_end_bench.md)

Dataset records (Zenodo):

- FernandinaSenDT128 — <https://zenodo.org/records/3952953>
- GalapagosSenDT128 — <https://zenodo.org/records/4743058>
- SanFranBaySenD42 — <https://zenodo.org/records/15814132>
- KujuAlosAT422F650 — <https://zenodo.org/records/3952917>
- SanFranSenDT42 — <https://zenodo.org/records/4265413>

s-sasaki-earthsea-wizard:
@sourcery-ai Please review this PR.

s-sasaki-earthsea-wizard:
Hi @huchangyang — thanks for taking this on, and happy to make whatever changes help. Summarizing where I've landed on the latest Sourcery pass (2026-06-04) and flagging the one place I'd like your call before expanding the test suite.

## Clear-cut — I'll fix this:

- The inline comment above the GPU branch in `ifgram_inversion.py` is stale: it still says rank-deficient pixels are "not handled" and "NaN/Inf will propagate downstream." The Cholesky path (`_solve_cholesky`) actually detects rank deficiency via `cholesky_ex` info codes and zeros those solutions. I'll update the comment to match.

## False positive — no action:

- Sourcery reports `requirements-gpu.txt` as empty. On the PR branch it does declare `torch>=2.11`; the bot's diff renderer seems to have missed the file body. Noting it so it doesn't cause confusion.

## Test coverage:

One thing worth context: CircleCI installs `requirements.txt` (+ gdal, pytest) only — not the `[gpu]` extra — so `torch` is absent on CI, and `test_ifgram_inversion_gpu.py` (guarded by `importorskip("torch")`) is skipped there entirely. The numerical-equivalence tests genuinely need a CUDA GPU, so I think keeping them gated/skipped on CI is correct — I'd rather skip honestly than monkeypatch a fake GPU and run the solver on CPU, which wouldn't validate the real path.

Given that, here's how I'd handle Sourcery's test suggestions:

1. **Solver-availability gating** — I'll add this, but as an *import-safe* test (no `importorskip`) that runs natively on CI's torch-absent state: it asserts `is_solver_available('torch')` is `False` and `_get_torch_device('torch')` raises a clear `ImportError` with the install hint — i.e. exactly the path a CPU-only user hits with `--solver torch`. This covers the gating logic on CI without faking hardware.
2. **Rank-deficient-pixel path** (warning + zeroed solution + finite `inv_quality`) and 3. **`min_redundancy` early-return branch** — both look like legitimate coverage gaps to me. The catch is they exercise the GPU solver, so they'd only run on a CUDA machine and stay skipped on CI like the rest of the equivalence suite.

Before I expand the suite: does adding (2) and (3) match the level of coverage you'd want here, given they only execute on CUDA hardware? And is there anything else you'd like addressed in this PR?

s-sasaki-earthsea-wizard:
Hi @yunjunz @huchangyang — a gentle follow-up on this one (no rush, I know the repo has been busy lately).

I'd like to make this as low-cost as possible to evaluate. The GPU path is fully opt-in: `mintpy.networkInversion.solver` defaults to `auto → cpu`, the CPU code path is byte-for-byte unchanged, and `torch` is an optional `[gpu]` extra rather than a new required dependency — so existing installs and CI are unaffected. The goal is simply to give users with an NVIDIA GPU a much faster `invert_network` (5–90× across the 5 benchmarked scenes), so time-series re-processing and parameter iteration become quicker, with nothing changed for CPU users.

If there are any concerns — maintenance burden, dependency footprint, scope, or how GPU support fits the project long-term — I'd be glad to talk them through, trim the scope, or rebase as needed. I'm also happy to maintain the GPU path and take point on any GPU-related issues after merge.

Is there anything I can do to make review easier?

yunjunz:
Thank you @s-sasaki-earthsea-wizard for the awesome PR! We are busy with the ISCE+ short course preparation at the beginning of next month. We will get back to the reviewing/testing of this PR after that.

s-sasaki-earthsea-wizard:
Thanks @yunjunz — that's great to hear, and no rush at all. 
Good luck with the ISCE+ short course prep!

I'll keep the branch rebased on `main` so it stays ready whenever you get to it, 
and I'm happy to answer questions or adjust scope when the time comes. 
Looking forward to it.

s-sasaki-earthsea-wizard:
Rebased onto the current `main` (`9ede611`) — no conflicts, and the three original commits are unchanged (verified with `git range-diff`). Local gates re-run after the rebase: `pre-commit run --all-files` clean, GPU unit tests 6/6 pass on CUDA (RTX 5080).

This push also includes the comment fix promised in the review discussion above: the inline comment at the GPU dispatch branch now describes the actual rank-deficient handling in `_solve_cholesky` (zero-fill with a warning) instead of the stale NaN/Inf note.

s-sasaki-earthsea-wizard:
Rebased onto the current `main` (`9ede611`) — no conflicts, and the three original commits are unchanged (verified with `git range-diff`). Local gates re-run after the rebase: `pre-commit run --all-files` clean, GPU unit tests 6/6 pass on CUDA (RTX 5080).

This push also includes the comment fix promised in the review discussion above: the inline comment at the GPU dispatch branch now describes the actual rank-deficient handling in `_solve_cholesky` (zero-fill with a warning) instead of the stale NaN/Inf note.

